PR #11, opened 2020-03-12
change the default controller image name

bitsf:

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -20,7 +20,7 @@ spec:
         - /manager
         args:
         - --enable-leader-election
-        image: controller:latest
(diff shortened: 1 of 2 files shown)


config/manager/manager.yaml
```diff
@@ -20,8 +20,9 @@ spec:
         - /manager
         args:
         - --enable-leader-election
-        image: controller:latest
+        image: goharbor/harbor-operator:dev
         name: manager
+        imagePullPolicy: IfNotPresent
```
jMonsinjon: Not necessary, it's the default behavior for tags != 'latest'.
I think we should use default strategy.

bitsf: fixed

PR #13, opened 2020-03-13
GitHub actions: ci-cd

### LINKED ISSUE #3: Enable CI/CD pipeline with github action
Cover:
  - CODE (lint, vet, race )
  - UT
  - Verify if it can work well in cluster (kind/minikube etc.）

holyhope:
ci(lint) Add linters github actions. Fix #3


.github/workflows/build.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build latest
+      run: docker build -t goharbor.io/harbor-operator:${{ github.sha }}
+    # TODO Push image to the registry
+    - run: docker tag goharbor.io/harbor-operator:${{ github.sha }} goharbor.io/harbor-operator:latest
```
bitsf: our repo is https://hub.docker.com/u/goharbor

bitsf: I think we should use tag dev for daily build, and version 0.5.0 for future release

holyhope: Ok for version 0.5.0, I suggest to use `goharbor/harbor-operator:`[`${{ github.ref }}`](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context), WDYT?
About `:dev`, when you are talking about *daily build*, you are talking about `master`? If yes, IMO `:latest` is better. It forces us to have a working `master` branch and make life easier for other people. Why do you want `:dev` tag? And so, what would be the `:latest`, probably the latest stable release?

bitsf: But if using latest, we have to push the image to dockerhub, because it is not set `imagePullPolicy: IfNotPresent`.
And the reason why I also suggest using dev is it also used by images in goharbor/harbor when compiling master branch, and developer can make install directly on self environment.

PR #14, opened 2020-03-14
use harbor-sample-certificate

bitsf:
should use harbor-sample-certificate instead of public-certificate

### DIFF
--- a/config/samples/certificate.yaml
+++ b/config/samples/certificate.yaml
@@ -10,7 +10,7 @@ kind: Certificate
 metadata:
   name: harbor-sample-public
 spec:
-  secretName: public-certificate
+  secretName: harbor-sample-certificate
   dnsNames:
     - '{{ env.Getenv "LBAAS_DOMAIN" }}'
     - '{{ env.Getenv "NOTARY_DOMAIN" }}'


config/samples/certificate.yaml
```diff
@@ -10,7 +10,7 @@ kind: Certificate
 metadata:
   name: harbor-sample-public
 spec:
-  secretName: public-certificate
+  secretName: harbor-sample-certificate
```
holyhope: The goal of this Certificate resource is to generate the [`tlsSecretName` of harbor sample](https://github.com/goharbor/harbor-operator/blob/master/config/samples/goharbor_v1alpha1_harbor.yaml#L7).  
[./config/samples/goharbor_v1alpha1_harbor.yaml](https://github.com/goharbor/harbor-operator/blob/master/config/samples/goharbor_v1alpha1_harbor.yaml#L7) needs to be updated too.  
Such as other reference in documentation [./docs/installation.md](https://github.com/goharbor/harbor-operator/blob/master/docs/installation.md#deploy-the-sample)

bitsf:
hi @holyhope 
I just wonder how harbor-sample-certificate and harbor-sample-notary-certificate created, and is harbor-sample-certificate designed for harbor nginx https ?

image

holyhope:
These certificates are created by the controller and are for _internal_ use, it needs more documentation. I just created the issue #18.
Otherwise I suggest to rename `public-certificate` to `harbor-sample-public-certificate`, WDYT?

bitsf:
Sounds good.
I thought public-certificate is used by mistake.
Ss harbor-sample-certificate is used between different components, such as core and jobservice, right?

holyhope:
>  Ss harbor-sample-certificate is used between different components, such as core and jobservice, right?

That is it.

PR #22, opened 2020-03-19
feat(cd) Release and tests

holyhope:
Push docker images depending of the github event type.
Cf <https://help.github.com/en/actions/reference/events-that-trigger-workflows>


holyhope:
Some adjustment needed: `KUBECONFIG` not found during `$ make test`

holyhope:
Not sure about the `CODEOWNERS` file.
The goal is to add default reviewer, but maybe there is too many people in that group.
WDYT?
Otherwise, who would be reviewer on this repo?

holyhope:
cc @goharbor/harbor-operator-dev

.github/workflows/tests.yml
```diff
@@ -0,0 +1,162 @@
...
+  go-tests:
+    runs-on: ubuntu-latest
+    name: K8S v${{ matrix.k8sVersion }} (CM v${{ matrix.certManager }})
+
+    strategy:
+      matrix:
+        # https://github.com/jetstack/cert-manager/branches/active
+        certManager: ["0.12", "0.13", "0.14", "0.15"]
+        # https://snapcraft.io/microk8s
+        k8sVersion: ["1.15", "1.16", "1.17"]
```
bitsf: I think we don't need test so many matrix

bitsf:
> CODEOWNERS

it seems we don't have such CODEOWNERS in goharbor/harbor, so I think this is not neccessary

.github/workflows/tests.yml
holyhope: I had issue with 1.14 when Ingresses were in `extv1` and not yet `netv1`.
IMO it is important to have the compatibility matrix. To be honest we have cluster still running on 1.15...
Moreover it does not cost so much to test on many versions thanks to that matrix syntax.
We could imagine to run 1.15 and 1.16 only on Pull Request? (Not sure how to do that properly)

holyhope:
> it seems we don't have such `CODEOWNERS` in `goharbor/harbor`, so I think this is not necessary

True.
I added this file because on this pull request there was no default reviewers. In order to not add people manually, I suggest that we add CODEOWNERS file. But probably overkill at this moment.
How can I add default reviewer on the project without owners?

bitsf:
> > it seems we don't have such `CODEOWNERS` in `goharbor/harbor`, so I think this is not necessary
> 
> True.
> I added this file because on this pull request there was no default reviewers. In order to not add people manually, I suggest that we add CODEOWNERS file. But probably overkill at this moment.
> How can I add default reviewer on the project without owners?

@steven-zou @reasonerjt Do you have any suggestion about this

.github/workflows/tests.yml
bitsf: I just feel the checks list is a little long with 43 items.
But it's OK if you really find it would helps.

PR #35, opened 2020-03-27
Dev/18 docs certificates role

holyhope:


holyhope:
I had to generate plantuml files only on master, otherwise it adds commits on the branch and does not re-trigger the workflow. The consequence is that we did not have build log in PR...
But now ready to review.
(Shall we add the plantUml generation in github action so?)

PR #38, opened 2020-04-01
feat(cicd) enable goreleaser with github action

### LINKED ISSUE #12: Introduce gorelease tool into this project to publish the coming releases easily

bitsf:
fix #12 
Use goreleaser to create a draft release when push a tag automatically.
The release will include the changelog generated by commit title, compiled binaries, some assets such as license and readme, and checksum file.

release
image


.github/workflows/tests.yml
```diff
@@ -2,6 +2,8 @@ name: Tests
 
 on:
   push:
+    branches:
```
holyhope, line 5: Are these changes really useful?

.github/workflows/goreleaser.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Unshallow
+        run: git fetch --prune --unshallow
```
holyhope, line 15:
```suggestion
        # required for the changelog to work correctly
        run: git fetch --prune --unshallow
```

.gitignore
```diff
@@ -29,3 +29,6 @@ pkged.go
 
 ## Mac file
 .DS_Store
+
+## goreleaser
+dist/
```
holyhope, line 34: Can you please add these line to `.dockerignore`?

.goreleaser.yml
```diff
@@ -0,0 +1,29 @@
+# This is an example goreleaser.yaml file with some sane defaults.
+# Make sure to check the documentation at http://goreleaser.com
+before:
```
holyhope, line 3: Since goreleaser config file is commited, can you suggest a way to install it?
Maybe a *goreleaser* target in `Makefile`?

.github/workflows/tests.yml
bitsf: This change makes the test not run again when doing tag

.goreleaser.yml
bitsf: I add goleaser task in Makefile, but not sure how to make it much useful, as the default install dir is ./bin/goreleaser

holyhope: Fair enough! ;-)

Makefile
```diff
@@ -277,3 +277,9 @@ HELM=helm-not-found
 else
 HELM=$(shell which helm)
 endif
+
+# find or download goreleaser
+goreleaser:
```
holyhope, line 282: Can you use the same way to install tools than for kubebuilder for example?

```make
goreleaser:
ifeq (, $(shell which goreleaser))
	# https://kubebuilder.io/quick-start.html#installation
	curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh \
		| sh -s v0.129.0
	mv ./bin/goreleaser "$(GORELEASER)"
GORELEASER=$(GOBIN)/kubebuilder
else
GORELEASER=$(shell which goreleaser)
endif
```

holyhope: Then it would be nice to have a target `release` or `gorelease`.
```make
release:
	$(GORELEASER) ...
```

.goreleaser.yml
```diff
@@ -0,0 +1,27 @@
...
+    - make generate
+    - git status
+builds:
+- env:
+  - CGO_ENABLED=0
+  binary: manager
+  goos:
+  - linux
+  goarch:
+  - amd64
```
steven-zou, line 16: Only `amd` supported?

.goreleaser.yml
```diff
@@ -0,0 +1,27 @@
+before:
+  hooks:
+    # You may remove this if you don't use go modules.
+    #- go mod download
+    # you may remove this if you don't need go generate
+    #- go generate ./...
+    - make generate
+    - git status
+builds:
```
steven-zou, line 9: No docker image option?

bitsf: We already have docker push in another action

.goreleaser.yml
bitsf: harbor doesn't support other arch now, so there's no need to release for others too

Makefile
bitsf: fixed

PR #52, opened 2020-04-18
fix(cd) Use only version part from ${{ github.ref }} as docker tag

holyhope:


.github/workflows/release.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - run: echo -n '${{ github.ref }}' | sed 's|refs/tags/||' > ./target/TAG_NAME
```
bitsf, line 12: I think you should `mkdir target` first

PR #56, opened 2020-06-10
Some documentation fixes

jsuchome:
Here are some minor fixes for documentation. 
However one is also affecting the Makefile, so please make sure it does not break your dev setups.

Signed-off-by: Jiří Suchomel <jiri.suchomel@suse.com>


holyhope:
Thanks a lot @jsuchome for your contribution

PR #64, opened 2020-07-19
redirect: disable: false` should be exposed at the harbor CR spec

### LINKED ISSUE #63: The registry configuration `redirect: disable: false` should be exposed at the harbor CR spec
Found issues in the cluster-operator: https://github.com/goharbor/harbor-cluster-operator/issues/118

registry configuration references: https://docs.docker.com/registry/configuration/

similar configurations in helm chart:  https://github.com/goharbor/harbor-helm/blob/fad6918c0fe3ca916d0948b98d433f6bffc96035/values.yaml#L243

soulseen:
Signed-off-by: xiaoyang zhu <zhuxiaoyang1996@gmail.com>

fix #63

### DIFF
--- a/api/v1alpha1/harbor_types.go
+++ b/api/v1alpha1/harbor_types.go
@@ -147,6 +147,9 @@ type RegistryComponent struct {
 
 	// +optional
 	CacheSecret string `json:"cacheSecret,omitempty"`
+
+	// +optional
+	DisableRedirect bool `json:"disableRedirect,omitempty"`
 }
 
 type RegistryControllerComponent struct {
--- a/assets/templates/registry/config.yaml
+++ b/assets/templates/registry/config.yaml
@@ -76,7 +76,7 @@ storage:
   delete:
     enabled: true
   redirect:
-    disable: false
+    disable: {{ env.Getenv "STORAGE_DISABLE_REDIRECT" }}
   cache:
     blobdescriptor: {{ if gt ( len $redisUrl ) 0 -}} redis {{- else -}} inmemory {{- end }}
   maintenance:
--- a/controllers/harbor/components/registry/deployments.go
+++ b/controllers/harbor/components/registry/deployments.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"strconv"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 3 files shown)


steven-zou:
@holyhope 

Would you please take a look at this PR? Thanks!

assets/templates/registry/config.yaml
```diff
@@ -74,7 +74,7 @@ compatibility:
 
 storage:
   delete:
-    enabled: true
+    enabled: {{ env.Getenv "STORRAGE_REDIRECT" }}
```
holyhope, line 77: Typo here

controllers/harbor/components/registry/deployments.go
```diff
@@ -149,6 +150,10 @@ func (r *Registry) GetDeployments(ctx context.Context) []*appsv1.Deployment { //
 									},
 								},
 								Env: []corev1.EnvVar{
+									{
+										Name: "STORRAGE_REDIRECT",
```
holyhope, line 154: Same typo here

api/v1alpha1/harbor_types.go
```diff
@@ -147,6 +147,10 @@ type RegistryComponent struct {
 
 	// +optional
 	CacheSecret string `json:"cacheSecret,omitempty"`
+
+	// +optional
+	Redirect bool `json:"redirect,omitempty"`
```
holyhope, line 152: This will change the default value from true to false

controllers/harbor/components/registry/deployments.go
Ghostbaby, line 154:
```suggestion
										Name: "STORAGE_REDIRECT",
```

api/v1alpha1/harbor_types.go
soulseen: yes, the default value is false, and the user can change it by the harbor CR spec.

steven-zou: @holyhope 

It seems @soulseen is correct, the previous default value is `false`.

Check the template file `assets/templates/registry/config.yaml`.

```yaml
redirect:	  redirect:
    disable: false
```

steven-zou:
@holyhope 

@soulseen has updated this PR, please review it again. Thanks a lot!

api/v1alpha1/harbor_types.go
holyhope: What does `.Spec.Redirect == true` means? Should `redirect: false` or `redirect: true`?
You are right default value for `.Spec.Redirect` is `false`. But I though that would mean `disable: true`.
What do you think about renaming it `DisableRedirect`?

```suggestion
	DisableRedirect bool `json:"disableRedirect,omitempty"`
```

soulseen: good point.

steven-zou:
@holyhope 

Renaming it to `DisableRedirect ` does make sense. Disabling the redirect can avoid the docker client directly connecting to the storage. That case may not suitable for some kinds of storages such as minIO etc..

@soulseen I think we can rename it.

soulseen:
I renamed it, please review it again thanks. @steven-zou @holyhope

steven-zou:
@holyhope 

We update the code following your comments, I'll merge this PR later and create a tag 0.5.1 to release a new patch version today.

PR #66, opened 2020-07-21
fix(ingress): resolve 413(Too Large Entity) error when push large image.

### LINKED ISSUE #63: The registry configuration `redirect: disable: false` should be exposed at the harbor CR spec

wangcanfengxs:
add annotation to unlimit the size of http request body.

CLOSES #63

### DIFF
--- a/controllers/harbor/components/registry/ingresses.go
+++ b/controllers/harbor/components/registry/ingresses.go
@@ -44,6 +44,10 @@ func (r *Registry) GetIngresses(ctx context.Context) []*netv1.Ingress { // nolin
 					"harbor":   harborName,
 					"operator": operatorName,
 				},
+				Annotations: map[string]string{
+					// resolve 413(Too Large Entity) error when push large image. It only works for NGINX ingress.
+					"nginx.ingress.kubernetes.io/proxy-body-size": "0",
+				},
 			},
 			Spec: netv1.IngressSpec{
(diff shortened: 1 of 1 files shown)


steven-zou:
@holyhope 

Could you please take a look? Thanks!

steven-zou:
@holyhope 

It seems the failed cases need to be revised, some configurations are not fresh now? Any suggestions?

holyhope:
> It seems the failed cases need to be revised, some configurations are not fresh now? Any suggestions?

Fyi I configured the ingress controller itself to have this behaviour. So tests will pass when installing the ingress controller with the make target. We could remove this helm value from the make target?

steven-zou:
> > It seems the failed cases need to be revised, some configurations are not fresh now? Any suggestions?
> 
> Fyi I configured the ingress controller itself to have this behaviour. So tests will pass when installing the ingress controller with the make target. We could remove this helm value from the make target?

@holyhope 

What's the purpose for introducing this make target?  What functions do you want to use this make target to test?

PR #92, opened 2020-08-29
Fix/clair updater and ingress cert

steven-zou:
- add cert manager related annotations to the ingresses (core and notary)
- add host info to the ingress TLS part
- remove REGISTRY_HTTP_HOST ENV VAR from registry deployment
- update clair updater interval to 1h to enable CVE DB updating process
- add csrf-key environment variable. use 32 bit random string.
- fix missing `_REDIS_URL` env of harbror-core
- fix core wrong cacheSecret

Signed-off-by: Steven Zou szou@vmware.com

### DIFF
--- a/api/v1alpha1/harbor_secret_format.go
+++ b/api/v1alpha1/harbor_secret_format.go
@@ -37,6 +37,8 @@ const (
 	HarborCoreDatabaseNameKey     = "database"
 	HarborCoreDatabaseUserKey     = "username"
 	HarborCoreDatabasePasswordKey = "password"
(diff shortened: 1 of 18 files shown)


steven-zou:
@holyhope 

It seems you reset the start point of 0.5.1 branch. Some of the originally merged commits are out of the scope.

Anyway, all the changes for running the operator are put into this PR.

steven-zou:
@holyhope 

I‘d like to invite you to give us(operator developers) an overall introduction about the new harbor operator next week. What do you think?

steven-zou:
Use #98

PR #98, opened 2020-09-01
Fix/clair updater and ingress cert

steven-zou:
- add cert-manager related annotations to the ingresses (core and notary)
- add host info to the ingress TLS part
- remove `REGISTRY_HTTP_HOST ` ENV VAR from registry deployment
- update Clair updater interval to `1h` to enable CVE DB updating process

Signed-off-by: Steven Zou <szou@vmware.com>


steven-zou:
Pending too long and forcely merged

PR #108, opened 2020-09-11
feat(apis) Expose github token for Trivy

### LINKED ISSUE #93: Github token config used by trivy should be exposed at Harbor spec level
`GithubTokenRef string `json:"githubTokenRef,omitempty"``

holyhope:
Fix #93


controllers/goharbor/harbor/github.go
```diff
@@ -2,54 +2,28 @@ package harbor
 
 import (
 	"github.com/ovh/configstore"
```
steven-zou, line 4: @holyhope 

Besides `File`, `ENV`, and `Memory` providers, what other providers we have for `github.com/ovh/configstore`? Especially in K8s env?

steven-zou: With `configstore`, I can change my configurations in place? Does that mean some config options like `GitHubToken` can be set by both Harbor spec and `configstore`?

steven-zou (commented):
commented

controllers/goharbor/harbor/github.go
holyhope: > Besides `File`, `ENV`, and `Memory` providers, what other providers we have for `github.com/ovh/configstore`? Especially in K8s env?

I think that there is no no such providers (but can implement one if needed). 
IMO there is (almost) no need for that: updating a configMap as volume in pod will propagate changes and configStore will hot-reload the file.
It is possible to imagine a configstore provider getting configs from k8s resources. It requires more access to the serviceAccount but there is no volumeMount needed.

> With `configstore`, I can change my configurations in place?

With the code I wrote in harbor-operator, configstore does not hot-reload githubtoken changes. Deployment needs a rollout. But the hot-reload can be implemented.

> Does that mean some config options like `GitHubToken` can be set by both Harbor spec and `configstore`?

Yes, that is right. Does it sound a good idea in your opinion? ATM consider it like a fallback.

steven-zou: > I think that there is no no such providers (but can implement one if needed).

I got the provider info from the readme of `github.com/ovh/configstore`.

steven-zou: >With the code I wrote in harbor-operator, configstore does not hot-reload githubtoken changes. Deployment needs a rollout. But the hot-reload can be implemented.

Although the hot-loaded is not supported. But via `configstore` I have more chances to configure my Harbor deployment when deploying Harbor, right?

steven-zou: >Yes, that is right. Does it sound a good idea in your opinion? ATM consider it like a fallback.

Yup.

@holyhope

steven-zou (approved):
LGTM

PR #109, opened 2020-09-11
feat(apis) Squash Harbor spec and remove HarborHelm1_4_0Spec

### LINKED ISSUE #97: Rename `HarborHelm1_4_0Spec` to a more general one?
I understand the current operator spec is referring to the harbor chart V1.4 templates.

My concern is anyway those two specs will be developed separately in the future. Some configuration may be introduced into the operator spec but not put into the chart template. And If the chart version is upgrading or templates are changed for fixing some bugs in the chart and maybe no influences to the operator, will the operator define a new spec object `HarborHelm1_5_0Spec` then?

holyhope:
Fix #97


steven-zou (approved):
lgtm

PR #110, opened 2020-09-14
Invert RBAC creation logic so the template matches the values

glitchcrab:
Signed-off-by: Simon Weald <glitchcrab-github@simonweald.com>


holyhope:
Thank you @glitchcrab
The Helm chart did not work, but #108 fixed it.

PR #89, opened 2020-08-28
fix: replace subpath mount by folder mount to use secret synchronization

holyhope:

PR #116, opened 2020-09-17
fix: replace subpath mount by folder mount to use secret synchronization

bitsf:
cherry-pick from https://github.com/goharbor/harbor-operator/pull/89


steven-zou (approved):
LGTM

PR #123, opened 2020-09-29
fix Makefile

bitsf:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 
 # Image URL to use all building/pushing image targets
 IMG ?= goharbor/harbor-operator:dev
+RELEASE_VERSION ?= 0.0.0-dev
 
 CONFIGURATION_FROM ?= env,file:$(CURDIR)/config-dev.yml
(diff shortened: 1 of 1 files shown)


holyhope (changes requested):
The chart needs to be generated before installing it.
We may add all chart files as dependency.
But I suggest to use the package:
```make
helm-install: helm $(CHARTS_DIRECTORY)/harbor-operator-$(RELEASE_VERSION).tgz
	$(HELM) upgrade --install $(CHART_RELEASE_NAME) $(CHARTS_DIRECTORY)/harbor-operator-$(RELEASE_VERSION).tgz \
		--set-string image.repository="$(shell echo $(IMG) | sed 's/:.*//')" \
		--set-string image.tag="$(shell echo $(IMG) | sed 's/.*://')" \
		--set-string harborClass='${CHART_HARBOR_CLASS}'
```

To make it work, it is necessary to move
```make
RELEASE_VERSION ?= 0.0.0-dev
```
at the top of the makefile.

PR #126, opened 2020-10-16
feat(arch) merge harbor-cluster-operator

bitsf:
merge harborcluster-operator and harbor-operator


.github/workflows/tests.yml
```diff
@@ -110,90 +140,127 @@ jobs:
...
+        kustomize edit add patch --path patch/github-token.yaml
+        kustomize edit set image goharbor/harbor-operator=${dockerImage}
+        kustomize build | kubectl apply -f -
+
+        # make helm-install IMG=${dockerImage}
+
+        if ! time kubectl -n ${operatorNamespace} wait --for=condition=Available deployment --all --timeout 300s; then
+          kubectl get all -n ${operatorNamespace}
+          exit 1
+        fi
```
holyhope, line 164: I suggest to keep the *helm install* way:
- IMO it is closer to the reality than `kustomize build | kubectl apply -f -`
- Multiple `kustomize edit ...` before running test looks like default values do not work.

.github/workflows/tests.yml
```diff
@@ -110,90 +140,127 @@ jobs:
...
-        for i in $(seq 1 6);do
+        IP=`hostname -I | awk '{print $1}'`
+        echo "IP=$IP" >> $GITHUB_ENV
+        CORE_HOST=core.$IP.nip.io
+        NOTARY_HOST=notary.$IP.nip.io
+        echo "CORE_HOST=$CORE_HOST" >> $GITHUB_ENV
+        echo "NOTARY_HOST=$NOTARY_HOST" >> $GITHUB_ENV
+        sed -i "s/core.harbor.domain/$CORE_HOST/g" manifests/samples/full_stack_other_fs.yaml
+        sed -i "s/notary.harbor.domain/$NOTARY_HOST/g" manifests/samples/full_stack_other_fs.yaml
+        sed -i "s/logLevel: info/logLevel: debug/g" manifests/samples/full_stack_other_fs.yaml
```
holyhope, line 177: Looks like a bit hacky.
The target `make sample` does not work anymore?
I suggest to add few parameters: `make sample CORE_HOST=$CORE_HOST NOTARY_HOST=$NOTARY_HOST`, wdyt?

.github/workflows/tests.yml
```diff
@@ -246,20 +313,9 @@ jobs:
...
-        - samples/database
         - samples/harbor
         - samples/harbor-full
-        - samples/jobservice
-        - samples/notary
-        - samples/notary-server
-        - samples/notary-signer
-        - samples/portal
-        - samples/registry
-        - samples/registryctl
```
holyhope, line 260: :-(

.golangci.yml
```diff
@@ -72,6 +66,12 @@ linters:
   - godox
   - paralleltest
   - wrapcheck
+  - errorlint
+  - goerr113
+  - gomnd
+  - nestif
+  - funlen
+  - gosec
```
holyhope, line 74: :-(
Maybe add an issue to re-enable major linters?
IMO `gosec` is pretty important.
Why did you disable it? I suggest to have punctual `// nolint:` comments.

Dockerfile
```diff
@@ -3,6 +3,6 @@
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY manager .
-USER nonroot:nonroot
+USER 65532:65532
```
holyhope, line 6: Please revert this change: https://github.com/goharbor/harbor-operator/pull/420#discussion_r569434764

Makefile
```diff
@@ -66,9 +66,12 @@ test: go-test go-dependencies-test
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 .PHONY: run
-run: go-generate certmanager $(TMPDIR)k8s-webhook-server/serving-certs/tls.crt
+run: go-generate vendor certmanager $(TMPDIR)k8s-webhook-server/serving-certs/tls.crt
```
holyhope, line 69: Do you want to re-introduce the vendor folder?

Makefile
```diff
@@ -201,7 +205,7 @@ dist/harbor-operator_linux_amd64/manager:
 # Run go linters
 .PHONY: go-lint
 go-lint: golangci-lint vet go-generate
-	$(GOLANGCI_LINT) run --verbose
+	$(GOLANGCI_LINT) run --verbose --max-same-issues 0 --sort-results -D wrapcheck -D exhaustivestruct -D errorlint -D goerr113 -D gomnd -D nestif -D funlen -D gosec
```
holyhope, line 208: Some linters are already disabled in `.golangci.yml`, please do not disable any linter in command line.

apis/goharbor.io/v1alpha2/harbor_types.go
```diff
@@ -56,8 +58,10 @@ type HarborSpec struct {
 	// +kubebuilder:validation:Optional
 	InternalTLS HarborInternalTLSSpec `json:"internalTLS"`
 
-	// +kubebuilder:validation:Required
-	ImageChartStorage HarborStorageImageChartStorageSpec `json:"imageChartStorage"`
+	// Skip OpenAPI schema validation
+	// Use validating webhook to do verification (field required)
+	// +kubebuilder:validation:Optional
+	ImageChartStorage *HarborStorageImageChartStorageSpec `json:"imageChartStorage,omitempty"`
```
holyhope, line 64: Please, do not edit an existing apis version if it has been released.
I suggested to @steven-zou to implements those changes in `v1alpha3`.
About the new CRD cluster, do as you wish: keep it in v1alpha2 or move it to v1alpha3 (I suggest to move it at the same time).

Be careful about keep versions constant as much as possible once released.

PS: I suggest to keep about 3 versions on the master branch, so here keep the v1alpha2 as it is and add the v1alpha3 as needed.

controllers/goharbor/configuration/reconciler.go
```diff
@@ -0,0 +1,258 @@
+package configuration
```
holyhope, line 1: Really not sure about that one.
Talked about this to @steven-zou some weeks ago.
Using a `ConfigMap` to configure Harbor Custom Resource deny the possibility to have a schema validation.
So the validation (even the simpliest) should be done through a webhook.
Moreover config value are no more strongly typed.
There is no automatic client generation, documentation, ...
There is no built list of expected values.

TLDR; I suggest to move all keys expected in this config to the right CRD (Harbor Core? I am not even sure that it is the core CRD or not).

holyhope (changes requested):

holyhope:
There is some design/architecture pending questions, but you people did a really great job!
Thank you all.

.github/workflows/tests.yml
bitsf: `make sample` is still using old Harbor CR, not using HarborCluster CR yet, correctly we are focus on testing with yamls in folder manifests

Makefile
bitsf: ok, will remove it

Makefile
bitsf: ok

.github/workflows/tests.yml
bitsf: there're too many check jobs to watch, so I'd like to do some simplify, especially these samples tests are already covered by `samples/harbor-full`

apis/goharbor.io/v1alpha2/harbor_types.go
steven-zou: @holyhope 

CRD version will be upgraded to v1alpha3, we have an issue https://github.com/goharbor/harbor-operator/issues/397 tracking.

But I want to clarify here, to reduce future maintenance work, I'd like to say the upcoming major release 1.0 should be the start point for considering multiple CRD versions or CRD upgrading cases. Actually, the previous 0.5 and 0.6 have totally different architecture and design. And for those two versions, we also did not announce them as production-ready releases. We'll not consider any upgrading supports for those two 0.x versions.

controllers/goharbor/configuration/reconciler.go
steven-zou: @holyhope 

CRD may be a better way, but NOT 100% perfect. I have created an issue to track the work.https://github.com/goharbor/harbor-operator/issues/444.

So far, as we're focusing on fixing the stability related issues, we don't have enough resources to do the enhancements. Let's keep it as it is now (at least it's usable). If we release resources later, then we can start to have a try of doing enhancement.

holyhope (commented):
Reviewed 65 over 221 files.
Working on it :-)

controllers/goharbor/core/deployments.go
```diff
@@ -31,6 +39,7 @@ const (
 	InternalCertificatesVolumeName        = "internal-certificates"
 	InternalCertificateAuthorityDirectory = "/harbor_cust_cert"
 	InternalCertificatesPath              = ConfigPath + "/ssl"
+	PermittedRegistryTypesForProxyCache   = "docker-hub,harbor"
```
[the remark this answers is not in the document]
steven-zou: @bitsf let's fix this.

controllers/goharbor/core/core.go
```diff
@@ -92,9 +90,8 @@ func New(ctx context.Context, name string, configStore *configstore.Store) (comm
 		configTemplatePath = DefaultConfigTemplatePath
 	}
 
-	r := &Reconciler{
-		configError: config.ErrNotReady,
-	}
+	r := &Reconciler{}
+	r.configError = config.ErrNotReady
```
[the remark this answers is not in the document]
steven-zou: I think no function change, just a different code style preference ^_^

heww (approved):

steven-zou (approved):
Let's merge this develop branch into master from now on.

PR #130, opened 2020-10-23
feat(ci) fix helm-install target

### LINKED ISSUE #129: dependent Make taget `helm-install: helm chart` is not found in Make file
`helm chart` it should be `helm-generate`

bitsf:
fix #129 


steven-zou (approved):
approved

PR #139, opened 2020-10-26
fix(core) admin password wrong

### LINKED ISSUE #138: admin secret is not successfully injected into harbor core service
provide admin secret in the deployment manifest, however, the CR core uses its own generated admin int secret as the harbor admin secret, it breaks the injects of admin secret from upper CR harbor.

bitsf:
fix #138 
admin password set in config/samples/harbor/kustomization.yaml:22 is not pass to harbor core


steven-zou (approved):
LGTM

PR #142, opened 2020-10-27
enhance CI testing

bitsf:
add ci cases
1. install harbor-operator
2. install harbor
3. curl api
4. docker push

microk8s 1.17 doesn't support registry with ipv6

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,7 +61,6 @@ jobs:
     runs-on: ubuntu-latest
     name: K8S v${{ matrix.k8sVersion }} (CM v${{ matrix.certManager }})
     env:
-      KUBECONFIG: /tmp/kubeconfig-microk8s.yaml
       USE_EXISTING_CLUSTER: true
 
     strategy:
@@ -75,6 +74,7 @@ jobs:
 
         # https://snapcraft.io/microk8s
         k8sVersion:
+        - "1.17"
         - "1.18"
         - "1.19"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -90,30 +90,63 @@ jobs:
...
+        set -ex
+        # make helm-install
+        sed -i '1inamespace: harbor-op' config/default/kustomization.yaml
+        kubectl create ns harbor-op
+        kustomize build config/default | kubectl apply -f -
+        kubectl wait --for=condition=Ready pod --all --timeout 100s -n harbor-op
+
+    - name: install harbor
+      run: |
+        export GITHUB_USER=xxx
```
steven-zou, line 126:
```suggestion
     
```

.github/workflows/tests.yml
```diff
@@ -90,30 +90,63 @@ jobs:
...
-        export KUBECONFIG="$(realpath "$(mktemp kubeconfig.XXXXX)")"
-        export USE_EXISTING_CLUSTER="true"
+        make go-test
 
-        sudo microk8s.config > "${KUBECONFIG}"
+    - name: install harbor-operator
+      run: |
+        set -ex
+        # make helm-install
+        sed -i '1inamespace: harbor-op' config/default/kustomization.yaml
```
steven-zou, line 119: If we add the default namespace to Kustomization file, we may not need to do `sed` anymore, right?

steven-zou (commented):
commented

.github/workflows/tests.yml
bitsf: fix

.github/workflows/tests.yml
```diff
@@ -60,60 +60,124 @@ jobs:
...
+        for i in $(seq 1 6);do
+          sleep 30
+          echo $i
+          kubectl get all
+        done
+        if ! time kubectl wait --for=condition=Ready pod --all --timeout 600s ;then
+          echo install harbor failed
+          kubectl get all
+
+          for n in $(kubectl get po |grep -v Running|grep -v NAME|awk '{print $1}');do
```
steven-zou, line 151: @bitsf 

I don't think we need to describe each pod here. We can find the unready pods and print out the log  by calling kubectl logs <POD>

steven-zou (commented):
commented

.github/workflows/tests.yml
bitsf: it has `grep -v Running`

steven-zou (approved):
LGTM

PR #144, opened 2020-10-27
fix(crd) add default namespace in kustomization.yaml

### LINKED ISSUE #136: add default namespace in the default root kustomization yaml file
or there will be an error occurred when running `kustomize build config/default`

bitsf:
add default namespace `harbor-operator` in `config/default/kustomization.yaml`
close #136

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -2,6 +2,8 @@
 #commonLabels:
 #  someName: someValue
 
+namespace: harbor-operator-ns
+
 bases:
 - ../crd
 - ../manager


config/default/kustomization.yaml
```diff
@@ -2,6 +2,8 @@
 #commonLabels:
 #  someName: someValue
 
+namespace: harbor-operator
```
steven-zou, line 5:
```suggestion
namespace: harbor-operator-ns
```

steven-zou (commented):
how about add -ns suffix

steven-zou (approved):
LGTM

PR #147, opened 2020-10-27
Update installation doc

### LINKED ISSUE #105: The cert manager version in the installation guide isn't correct
The installation guide says the minimal version of cert manager is 1.11: https://github.com/goharbor/harbor-operator/blob/master/docs/installation/installation.md#minimal
But the highest version of cert manager is 1.0.1 for now

bitsf:
Update some installation doc
close #105


steven-zou (approved):
lgtm

PR #148, opened 2020-10-27
Fix chart ingress enable

### LINKED ISSUE #133: The nginx ingress crtl is still deployed when setting the `ingress.enabled`=false
```
ingress:
  # ingress.enabled -- Whether to install ingress controller Helm chart
  enabled: false

```

bitsf:
close #133

Fix chart ingress enable value not working


steven-zou (approved):
LGTM

Issue #131, opened 2020-10-23
Too many replica set are created for registry-ctl component


steven-zou:
Run `make sample`

Then check the resource, you'll find more than one registry-ctl replica sets that with 0, 0.
```
replicaset.apps/sample-harbor-registryctl-589b8bb8df   1         1         1       2m16s
replicaset.apps/sample-harbor-registryctl-5bbf7d4889   0         0         0       2m36s
replicaset.apps/sample-harbor-registryctl-644bd8fdbf   0         0         0       2m47s
replicaset.apps/sample-harbor-registryctl-644d9549b6   0         0         0       2m50s
replicaset.apps/sample-harbor-registryctl-687bbd59cd   0         0         0       2m34s
replicaset.apps/sample-harbor-registryctl-6988fbbfbd   0         0         0       2m42s
replicaset.apps/sample-harbor-registryctl-7464d974bb   0         0         0       2m45s
replicaset.apps/sample-harbor-registryctl-74c6d4f8cc   0         0         0       2m35s
replicaset.apps/sample-harbor-registryctl-7569fb84f4   0         0         0       2m52s
replicaset.apps/sample-harbor-registryctl-79cc6ccbf    0         0         0       2m46s
replicaset.apps/sample-harbor-registryctl-c58b8b4d8    0         0         0       2m56s
```

Deployment is not ready:

```
NAME                                        READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/sample-harbor-registryctl   0/1     1            0           2d3h
```

steven-zou:
one day later, get more replicaset:

```
replicaset.apps/sample-harbor-registryctl-54fc7f564    0         0         0       2d2h
replicaset.apps/sample-harbor-registryctl-5656d8d6f4   0         0         0       2d2h
replicaset.apps/sample-harbor-registryctl-647f499c85   0         0         0       31h
replicaset.apps/sample-harbor-registryctl-64b77c84fc   0         0         0       8h
replicaset.apps/sample-harbor-registryctl-678597bc6b   1         1         0       112m
replicaset.apps/sample-harbor-registryctl-687bc446c9   0         0         0       27h
replicaset.apps/sample-harbor-registryctl-6b8ccdcb9    1         1         0       163m
replicaset.apps/sample-harbor-registryctl-6dbc6ff85b   0         0         0       21h
replicaset.apps/sample-harbor-registryctl-6fb4fdd49    0         0         0       2d2h
[...]

PR #169, opened 2020-10-30
1.0/controllers tests

holyhope:
A new design for test in order to have functional tests for every component (here is the draft for chartmuseum).


holyhope:
Tests are failing ATM. I found that there is a bug on resource dependency compute. May be related to #131.

PR #185, opened 2020-11-05
fix(bug) notary/trivy can't install with github token

### LINKED ISSUE #118: It seems Harbor CR does not update
Reproduce steps:

install harbor core components with `make sample`, harbor "sample" is deployed 

install full harbor with `make sample-harbor-full`, no more components are deployed.

expectations: the optional components like `notary`, `chartmuseum` and `trivy` should be deployed.

bitsf:
close #118
add github token injection for trivy component
remote github token for notary, and use internal logic to do migration


controllers/goharbor/chartmuseum/chartmuseum.go
```diff
@@ -23,7 +23,7 @@ const (
 	DefaultConfigTemplatePath = "/etc/harbor-operator/chartmuseum-config.yaml.tmpl"
 	ConfigTemplateKey         = "template-content"
 	ConfigImageKey            = "docker-image"
-	DefaultImage              = "goharbor/chartmuseum-photon:v2.0.0"
+	DefaultImage              = "ap.cicd.harbor.vmwarecna.net/proxy/goharbor/chartmuseum-photon:v2.0.0"
```
steven-zou, line 26: Is the registry publish for all the internet?

controllers/goharbor/harbor/notarysigner.go
```diff
@@ -328,6 +329,17 @@ func (r *Reconciler) GetDefaultNotarySignerMigrationSource(ctx context.Context,
 type NotarySignerMigrationSecret graph.Resource
 
 func (r *Reconciler) AddNotarySignerMigrationSecret(ctx context.Context, harbor *goharborv1alpha2.Harbor) (NotarySignerMigrationSecret, error) {
+	logger.Get(ctx).WithName("notaryserver").Info("AddNotarySignerMigrationSecret", "MigrationEnabled", *harbor.Spec.Notary.MigrationEnabled)
+
+	if !*harbor.Spec.Notary.MigrationEnabled {
+		return nil, nil
+	}
+
+	// do not use github now
```
steven-zou, line 338: Do we need this hardcoded code gate for disabling GitHub token reading?

controllers/goharbor/harbor/notaryserver.go
```diff
@@ -165,6 +167,17 @@ func (r *Reconciler) GetDefaultNotaryServerMigrationSource(ctx context.Context,
 type NotaryServerMigrationSecret graph.Resource
 
 func (r *Reconciler) AddNotaryServerMigrationSecret(ctx context.Context, harbor *goharborv1alpha2.Harbor) (NotaryServerMigrationSecret, error) {
+	logger.Get(ctx).WithName("notaryserver").Info("AddNotaryServerMigrationSecret", "MigrationEnabled", *harbor.Spec.Notary.MigrationEnabled)
+
+	if !*harbor.Spec.Notary.MigrationEnabled {
+		return nil, nil
+	}
+
+	// do not use github now
```
steven-zou, line 176: Do we need this hardcoded code gate for disabling GitHub token reading?

steven-zou (commented):
Let's have more discussions about this PR?

controllers/goharbor/harbor/trivy.go
```diff
@@ -75,7 +76,11 @@ func (r *Reconciler) GetTrivyUpdateSecret(ctx context.Context, harbor *goharborv
 
 	token, err := r.GetGithubToken(TrivyGithubCredentialsConfigKey)
 	if err != nil {
-		return nil, serrors.UnrecoverrableError(err, serrors.OperatorReason, "cannot get default github credentials")
+		if _, ok := err.(configstore.ErrItemNotFound); ok {
```
steven-zou, line 79: Why return nil, nil when it is not found?

controllers/goharbor/notaryserver/deployments.go
```diff
@@ -107,24 +103,41 @@ func (r *Reconciler) GetDeployment(ctx context.Context, notary *goharborv1alpha2
 	}
 
 	initContainers := []corev1.Container{}
+	migrateCmd := ""
+	migrationEnvs := []corev1.EnvVar{}
+
+	if notary.Spec.MigrationEnabled == nil || *notary.Spec.MigrationEnabled {
```
steven-zou, line 109: default is true?

steven-zou: not set means migration is enabled?

steven-zou (commented):
commented

controllers/goharbor/harbor/trivy.go
bitsf: for trivy, it doesn't have to have a github token

controllers/goharbor/notaryserver/deployments.go
bitsf: yes, the default is true, or fresh installation will be fail

controllers/goharbor/harbor/trivy.go
steven-zou: I don't think we should check it here. For the case of not setting, we should check the token secret ref in the CR spec.

bitsf: it will find another default token if not set inspec, and if neither is found, it will return ErrItemNotFound

steven-zou: Offline discussion with zm, let's keep the way used in the code

steven-zou (approved):
LGTM

PR #195, opened 2020-11-19
fix(chart) remove helm dependencies

### LINKED ISSUE #194: remove helm dependencies
remove section `dependencies` in `charts/harbor-operator/Chart.yaml`
because cert-manager and ingress are not suggested to install as part of the operator, they should be installed once in cluster level

bitsf:
close #194


steven-zou (approved):
LGTM

Issue #204, opened 2020-11-24
NPE:  ResourceManage in cache is null


wangcanfengxs:
E1124 10:11:06.385514   23340 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 452 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1936d40, 0x2c4de00)
	/root/harbor-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/root/harbor-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x82
panic(0x1936d40, 0x2c4de00)
	/root/go/src/runtime/panic.go:967 +0x166
github.com/goharbor/harbor-operator/pkg/cluster/cache.(*RedisController).Deploy(0xc00040ad80, 0xc000c23200, 0x1b6ddbb, 0x8, 0x0)
	/root/harbor-operator/pkg/cluster/cache/deploy.go:26 +0xc5
github.com/goharbor/harbor-operator/pkg/cluster/cache.(*RedisController).Apply(0xc00040ad80, 0x1e412c0, 0xc000931a10, 0xc000c23200, 0x0, 0x0, 0x0)
	/root/harbor-operator/pkg/cluster/cache/redis.go:62 +0x578
github.com/goharbor/harbor-operator/controllers/goharbor/harborcluster.(*Reconciler).AddResources(0xc0001d00c0, 0x1e412c0, 0xc000931a10, 0x1e74420, 0xc000c23200, 0x0, 0x0)
	/root/harbor-operator/controllers/goharbor/harborcluster/resources.go:28 +0x1c0
github.com/goharbor/harbor-operator/pkg/controller.(*Controller).Run(0xc0000a7620, 0x1e412c0, 0xc0009319e0, 0x1e74420, 0xc000c23200, 0x0, 0x0)
	/root/harbor-operator/pkg/controller/common.go:149 +0x1b5
github.com/goharbor/harbor-operator/pkg/controller.(*Controller).Reconcile(0xc0000a7620, 0xc0005aa589, 0x7, 0xc000307940, 0x14, 0xc000977d00, 0x0, 0x0, 0x0)
[...]

PR #206, opened 2020-11-24
fix(redis-controller): fix npe issue and some logic error

chlins:

Fix: https://github.com/goharbor/harbor-operator/issues/204

### DIFF
--- a/pkg/cluster/controllers/cache/redis.go
+++ b/pkg/cluster/controllers/cache/redis.go
@@ -47,13 +47,14 @@ type RedisController struct {
 	Scheme             *runtime.Scheme
 	RedisConnect       *RedisConnect
 	ResourceManager    ResourceManager
-	expectCR, actualCR *redisOp.RedisFailover
+	expectCR, actualCR runtime.Object
 }
 
 // Apply creates/updates/scales the resources, like kubernetes apply operation.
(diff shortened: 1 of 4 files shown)


Ghostbaby (approved):
lgtm

pkg/cluster/controllers/cache/redis.go
```diff
@@ -47,13 +47,14 @@ type RedisController struct {
 	Scheme             *runtime.Scheme
 	RedisConnect       *RedisConnect
 	ResourceManager    ResourceManager
-	expectCR, actualCR *redisOp.RedisFailover
+	expectCR, actualCR runtime.Object
```
steven-zou, line 50: Why use `runtime.Object` instead of `redisOp.RedisFailover`?

steven-zou (commented):
one comment

pkg/cluster/controllers/cache/redis.go
chlins: If use redisOp.RedisFailover, when controller gets actual cr from crd client, the type of crd client return if unstructed.Unstructed, which implements runtime.Object, so we can directly use rc.actualCR=${crdReturn}.

steven-zou: 👌

steven-zou (approved):
LGTM

PR #242, opened 2020-11-29
fix(notary) signer alias key should use upper case envrionment

### LINKED ISSUE #193: notary not function well
steps:
make sample-harbor-full
DOCKER_CONTENT_TRUST=1
DOCKER_CONTENT_TRUST_SERVER=https://notary.harbor.domain
docker push core.harbor.domain/library/busybox:test

failed in notarysigner
{"go.version":"go1.13.8","level":"error","msg":"CreateKey: failed to create key: failed to store key: expected env variable to not be empty: default","time":"2020-11-17T12:47:10Z"}

image
image
image

bitsf:
close #193 

the default alias key for notary signer should be upper case

reference https://github.com/theupdateframework/notary/blob/84287fd8df4f172c9a8289641cdfa355fc86989d/cmd/notary-signer/config.go#L88

### DIFF
--- a/controllers/goharbor/harbor/notarysigner.go
+++ b/controllers/goharbor/harbor/notarysigner.go
@@ -2,6 +2,7 @@ package harbor
 
 import (
 	"context"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


steven-zou (approved):
lgtm

PR #244, opened 2020-11-29
Fix notary not shown on UI

### LINKED ISSUE #243: notary signing info is not shown on GUI
steps:
make sample-harbor-full

image

There should be a signed marker on tag view
image

curl https://core.harbor.domain/api/v2.0/systeminfo
with_notary is false
image

bitsf:
close #243
notary is not shown on UI


steven-zou (approved):
LGTM

PR #247, opened 2020-11-30
fix(trivy) failed pull image

### LINKED ISSUE #246: trivy is not function well
install harbor with trivy enabled, and also setup the correct github token
but scan failed
image

bitsf:
close #246
trivy can't pull image with auth fail because it is not using internal auth url, because core_url doesn't have port

### DIFF
--- a/controllers/goharbor/core/deployments.go
+++ b/controllers/goharbor/core/deployments.go
@@ -119,9 +119,16 @@ func (r *Reconciler) GetDeployment(ctx context.Context, core *goharborv1alpha2.C
 		scheme = "https"
 	}
 
+	host := r.NormalizeName(ctx, core.GetName())
+	if core.Spec.Components.TLS.Enabled() {
+		host += ":443"
+	} else {
+		host += ":80"
+	}
+
 	coreLocalURL := (&url.URL{
 		Scheme: scheme,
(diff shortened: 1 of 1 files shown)


bitsf:
this issue is because harbor pr https://github.com/goharbor/harbor/pull/11755/files not handle trivy request well

controllers/goharbor/core/deployments.go
```diff
@@ -119,9 +119,16 @@ func (r *Reconciler) GetDeployment(ctx context.Context, core *goharborv1alpha2.C
 		scheme = "https"
 	}
 
+	host := r.NormalizeName(ctx, core.GetName())
+	if core.Spec.Components.TLS.Enabled() {
+		host += ":443"
+	} else {
+		host += ":80"
```
steven-zou, line 126: why do we need to set such default ports?

the protocol scheme can not resolve the problem?

steven-zou (commented):
I have questions about the change

controllers/goharbor/core/deployments.go
bitsf: because weiwei concat the port forcely with this PR
https://github.com/aquasecurity/harbor-scanner-trivy/issues/108
image

steven-zou (approved):
LGTM

PR #260, opened 2020-12-02
feat(api) upgrade ingress api from v1beta to v1

### LINKED ISSUE #217: api version of ingress should be upgraded to v1
Receive warning message when starting controller:

>networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress

bitsf:
feat(api) upgrade ingress api from v1beta to v1
close #217


steven-zou:
@bitsf 

It seems after this change, we can only support v1.19+ version, right?  We need to have a discussion offline.

steven-zou:
as discussed offline, let's support it when we supporting k8s v1.21+

PR #278, opened 2020-12-04
Fix/registry storage environments

holyhope:


steven-zou:
@holyhope 

Could you please check why only the V1.19 job is failed?

cc @bitsf

steven-zou (approved):
LGTM

bitsf (approved):

PR #288, opened 2020-12-07
Refactor/harbor ctrl

steven-zou:
refactor(ctrl):update harbor CR ctrl

- do comparison before the update
- refactor status objects
- introduce `equality.Semantic` to do k8s object spec comparison

refactor(reconciler):update the logic of cluster ctrl

- introduce a service manager to handle all the dependent services apply the process
- add service config getter func template for getting info to do a health check
- update the methods of cluster status handler

Signed-off-by: Steven Zou <szou@vmware.com>


bitsf (approved):

chlins (approved):
LGTM

PR #299, opened 2020-12-08
fix full_stack.yaml

bitsf:
close https://github.com/goharbor/harbor-cluster-operator/issues/235
some setting is not correct in sample full_stack.yaml


chlins (approved):
LGTM

Ghostbaby (approved):
lgtm

PR #321, opened 2020-12-11
fix ci for cluster-operator

bitsf:
change CI to use harborcluster


steven-zou (approved):
Approved.

Let's merge this PR now and switch storage from fs to minio when minio is totally fine later.

PR #322, opened 2020-12-11
update harbor version to 2.1.2

bitsf:


pkg/config/config.go
```diff
@@ -18,6 +18,7 @@ const (
 	DefaultConcurrentReconcile = 1
 	DefaultHarborClass         = ""
 	DefaultRegistry            = ""
+	DefaultHarborVersion       = "v2.1.2-rc1"
```
steven-zou, line 21: let's use the standard 2.1.2 as it has been released

steven-zou (commented):
commented

steven-zou (approved):
LGTM

PR #328, opened 2020-12-13
fix encrypt key length to 16

### LINKED ISSUE #326: fail to add harbor registry with credentials
{"errors":[{"code":"INTERNAL_SERVER_ERROR","message":"crypto/aes: invalid key size 128"}]}
image

bitsf:
close #326 
fix harbor-core encrypt key length to 16


steven-zou (approved):
LGTM

Issue #343, opened 2020-12-16
The registry can not work when redirect.disable is false.


chlins:
What can we help you?

![image](https://user-images.githubusercontent.com/31262637/102325342-febb7900-3fbd-11eb-93e3-7a3aecbc0fc0.png)

PR #346, opened 2020-12-17
fix: fix full stack certs lost minio domain and minio s3 enable v4auth

chlins:
Fix：https://github.com/goharbor/harbor-operator/issues/343


steven-zou (approved):
Approved

PR #373, opened 2021-01-02
support self-signed certificate and minio host

soulseen:
support self-signed certificate and minio host


apis/goharbor.io/v1alpha2/harborcluster_types.go
```diff
@@ -130,6 +131,12 @@ type MinIOSpec struct {
 	// Number of persistent volumes that will be attached per server
 	// +kubebuilder:validation:Minimum:=1
 	VolumesPerServer int32 `json:"volumesPerServer"`
+	// ExternalCertSecret allows a user to specify custom CA certificate, and private key. This is
+	// used for enabling TLS support on MinIO Pods.
+	// +kubebuilder:validation:Optional
+	ExternalCertSecret *minio.LocalCertificateReference `json:"externalCertSecret,omitempty"`
```
steven-zou, line 137: It seems `ExternalCertSecret` and `Host` should be set only when the` redirect.disable=false`, right? So, it may be better to put these options into the `Redirect StorageRedirectSpec` section as part of it?

apis/goharbor.io/v1alpha2/harborcluster_types.go
```diff
@@ -130,6 +131,12 @@ type MinIOSpec struct {
...
+	// +kubebuilder:validation:Optional
```
steven-zou, line 138: I think `Host` should be required when it needs to be exposed.

pkg/cluster/controllers/storage/provision.go
```diff
@@ -47,23 +47,28 @@ func (m *MinIOController) getMinIOProperties(minioInstance *minio.Tenant) (*goha
...
+		skipVerify = true
 	)
 
 	if !m.HarborCluster.Spec.InClusterStorage.MinIOSpec.Redirect.Disable {
-		scheme, endpoint, err = GetMinIOHostAndSchema(m.HarborCluster.Spec.ExternalURL)
-		if err != nil {
-			return nil, err
+		scheme = "https"
+		if minioHost = m.HarborCluster.Spec.InClusterStorage.MinIOSpec.Host; minioHost == "" {
+			scheme, minioHost, err = GetMinIOHostAndSchema(m.HarborCluster.Spec.ExternalURL)
```
steven-zou, line 62: I don't think we need to auto-generate the host from the `ExternalURL ` if the `host` is not set. See my above comment, the `host` should be required, not `optional`.

pkg/cluster/controllers/storage/provision.go
```diff
@@ -252,8 +257,9 @@ func (m *MinIOController) generateMinIOCR() *minio.Tenant {
 				Labels:      m.getLabels(),
 				Annotations: m.generateAnnotations(),
 			},
-			ServiceName: m.getServiceName(),
-			Image:       m.GetImage(),
+			ExternalCertSecret: m.HarborCluster.Spec.InClusterStorage.MinIOSpec.ExternalCertSecret,
+			ServiceName:        m.getServiceName(),
+			Image:              m.GetImage(),
```
steven-zou, line 262: Conflict with other merged PRs, please fix the conflicts before merging.

steven-zou (changes requested):
may need to do changes

apis/goharbor.io/v1alpha2/harborcluster_types.go
soulseen: In the case that the user does not need the exposed `Host`, the `Host` can be left unset, so maybe it should be `Optional`.

steven-zou: I think we should consider together with the above comment.

apis/goharbor.io/v1alpha2/harborcluster_types.go
```diff
@@ -113,11 +114,18 @@ type StorageRedirectSpec struct {
 	// Default is false
 	// +optional
 	Disable bool `json:"disable"`
```
steven-zou, line 116: `Disable ` => `Enable`(default value=true)

```yaml
redirect:
  enable: true
  route: # optional
     tls: tls_secret # optional
     host: my.minio.io # required
```

steven-zou (changes requested):
pls do changes

apis/goharbor.io/v1alpha2/harborcluster_types.go
heww: suggest using `expose` instead of `route`, because there is a `HarborExposeComponentSpec` in the types.

``` yaml
redirect:
   enable: true
   expose:
      tls: tls_secret
      ingress:
          host: my.minio.io
          controller: xxx
          annotations: xxx
```

steven-zou: 可以 @soulseen

soulseen: so we use `HarborExposeComponentSpec` directly, right? 

```yaml
Expose *HarborExposeIngressSpec `json:"expose,omitempty"`
```

soulseen:
```yaml
redirect:
   enable: true
   expose:
      tls: 
        certificateRef: tls_secret
      ingress:
          host: my.minio.io
          controller: xxx
          annotations: xxx
```

chlins (approved):
LGTM

apis/goharbor.io/v1alpha2/registry_types.go
```diff
@@ -708,6 +708,10 @@ type RegistryStorageDriverS3Spec struct {
 	// Skips TLS verification when the value is set to true.
 	SkipVerify bool `json:"skipverify"`
 
+	// +kubebuilder:validation:Required
```
heww, line 711: It should be optional.

apis/goharbor.io/v1alpha2/harborcluster_types.go
```diff
@@ -114,16 +114,18 @@ type Storage struct {
...
+	// +kubebuilder:validation:Optional
+	Expose *HarborExposeComponentSpec `json:"expose,omitempty"`
 }
 
 type MinIOSpec struct {
 	harbormetav1.ImageSpec `json:",inline"`
 
 	// Determine if the redirection of minio storage is disabled.
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
```
steven-zou, line 128:
```suggestion
	// +kubebuilder:validation:Optional
```

I think it's better to keep this item as an optional item.

But at the backend logic, we can keep the default behavior is `enable` the redirect if it is not set here.

steven-zou (commented):
Still has some comments

steven-zou (approved):
LGTM

PR #385, opened 2021-01-11
feat(loadblance) support to expose service via load balancer

### LINKED ISSUE #178: Support to expose service via load balancer
Currently the only option is ingress. Load balancer is useful for deploying on cloud

Ghostbaby:
support to expose service via load balancer

fix #178


apis/goharbor.io/v1alpha2/core_types.go
```diff
@@ -73,6 +73,9 @@ type CoreSpec struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Pattern="[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
 	CSRFKeyRef string `json:"csrfKeyRef"`
+
+	// +kubebuilder:validation:Optional
+	LoadBalancer bool `json:"loadbalancer,omitempty"`
```
steven-zou, line 78:
```suggestion
expose:
  core:
    loadBalancer:
      enable: true

```

steven-zou (changes requested):
Please do changes

apis/goharbor.io/v1alpha2/harbor_types.go
```diff
@@ -486,6 +486,9 @@ type HarborExposeComponentSpec struct {
 	Ingress *HarborExposeIngressSpec `json:"ingress,omitempty"`
 
 	// TODO Add supports to ClusterIP, LoadBalancer and NodePort by deploying the nginx component
+
+	// +kubebuilder:validation:Optional
+	LoadBalancer *HarborExposeLBSpec `json:"loadbalancer,omitempty"`
```
steven-zou, line 491: It seems we need to add some logic in the validation webhook to make sure the two types cannot be configured at the same time.

controllers/goharbor/core/services.go
```diff
@@ -38,3 +39,11 @@ func (r *Reconciler) GetService(ctx context.Context, core *goharborv1alpha2.Core
 		},
 	}, nil
 }
+
+func GetLBService(svc *corev1.Service) *corev1.Service {
```
steven-zou, line 43: Why do you introduce a new get service method? It seems you should do some modification to the original `reconciler.GetService()` method. The type of the getting service now is determined by the spec data, not directly use the default `ClusterIP` type.

steven-zou (changes requested):
Please review my comments and I think we need to do changes to the PR.

steven-zou (approved):
lgtm

PR #404, opened 2021-01-19
fix encrypt key length to 16

bitsf:
fix https://github.com/goharbor/harbor-operator/issues/326 for release-0.6.0


steven-zou (approved):
LGTM

PR #409, opened 2021-01-22
fix: inject github token from operator to trivy spec

heww:
1. Inject the github token to the trivy spec when the github token ref
is empty in the spec but github token exists in the operator.
2. Patch the GITHUB_TOKEN from the github actions to the operator in CI.
3. Use helm/kind-action to create k8s cluster.


bitsf (approved):

steven-zou (approved):
LGTM

PR #417, opened 2021-01-27
fix: correct GetImage when the respository is empty

heww:
Fix the GetImage function for cluster redis component


ywk253100 (approved):

steven-zou (approved):
LGTM

PR #420, opened 2021-01-27
fix: specify nonroot uid for manager

heww:
See https://github.com/kubernetes-sigs/kubebuilder/pull/1635 for more
info

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,6 @@
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY manager .
-USER nonroot:nonroot
+USER 65532:65532
 
 ENTRYPOINT ["/manager"]
--- a/charts/harbor-operator/README.md
+++ b/charts/harbor-operator/README.md
@@ -11,6 +11,7 @@ Deploy Harbor Operator
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` | Expects input structure as per specification <https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#affinity-v1-core> For example: `{   "nodeAffinity": {     "requiredDuringSchedulingIgnoredDuringExecution": {       "nodeSelectorTerms": [         {           "matchExpressions": [             {               "key": "foo.bar.com/role",               "operator": "In",               "values": [                 "master"               ]             }           ]         }       ]     }   } }` |
+| allowPrivilegeEscalation | bool | `false` | Allow privilege escalation for the controller Pods |
 | autoscaling.enabled | bool | `false` | Whether to enabled [Horizontal Pod Autoscaling](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) |
 | autoscaling.maxReplicas | int | `100` | Maximum conroller replicas |
 | autoscaling.minReplicas | int | `1` | Minimum conroller replicas |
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -3,6 +3,6 @@
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY manager .
-USER nonroot:nonroot
+USER 65532:65532
```
bitsf, line 6: 这个数是随便写的吗？咱们有没有统一的规范

heww: This user id is from https://github.com/kubernetes-sigs/kubebuilder/pull/1635 which is the part of the kubebuilder v3.

bitsf: 但是我看harbor这边用的都是10000
是不是我们应该统一

heww: This is the user id to run the harbor-operator itself, not the user id to run the harbor. 

I think we should keep it the same with kubebuilder because these files may be generated by the kubebuilder.

charts/harbor-operator/README.md
```diff
@@ -30,7 +30,6 @@ Deploy Harbor Operator
 | nodeSelector | object | `{}` | Expects input structure as per specification <https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#nodeselector-v1-core> For example: `[   {     "matchExpressions": [       {         "key": "kubernetes.io/e2e-az-name",         "operator": "In",         "values": [           "e2e-az1",           "e2e-az2"         ]       }     ]   } ]` |
 | podAnnotations | object | `{}` | Additional annotations to add to the controller Pods |
 | podLabels | object | `{}` | Additional labels to add to the controller Pods |
-| podSecurityContext | object | `{}` | Expects input structure as per specification <https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podsecuritycontext-v1-core> For example: `{   "fsGroup": 2000,   "runAsUser": 1000,   "runAsNonRoot": true }` |
```
steven-zou, line 33: Why did you remove this section?

heww: Because the operator manager default runs as a non-root user and runs as a user 65532, the `runAsUser` and `runAsNonRoot` is not needed now.

If the podSecurityContext not be removed, the customer needs to provide the podSecurityContext to run the operator in the restricted psp env.

steven-zou: If the users want to override the run user, they can also provide this section `podSecurityContext`, right? @heww

bitsf (approved):

charts/harbor-operator/README.md
bitsf: 他已经在dockerfile中写死user id了，这个参数估计没有了

bitsf: 他已经在dockerfile中写死user id了，这个参数估计没有了

steven-zou (approved):
LGTM

charts/harbor-operator/README.md
heww: I tested that the pod also runs successes when providing another user id in the security context.

Dockerfile
holyhope: I cannot read what @bitsf wrote,
But I agree with @heww : we should keep the `nonroot` user:
https://github.com/GoogleContainerTools/distroless/blob/master/examples/nonroot/BUILD#L17

Why using it here?
IMO The change should only be be applied to files under `config/`.

heww: This PR lets the harbor operator can run in the k8s cluster with PSP enabled.

The `config/manager/manager.yaml` was changed with `runAsUser: 65532`

holyhope: I got it, but the dockerfile does not require any changes.
The user id should be changed on runtime, not at buildtime.

The user `65532` does not exist in `gcr.io/distroless/static:nonroot` image.

PR #423, opened 2021-01-28
feat(harbocluster): fix health check err make reconcile long time delay

### LINKED ISSUE #424: There will be a non-reasonable delay between pods of Harbor CR creations and all services ready status
What can we help you?

wangcanfengxs:

CLOSE #424

### DIFF
--- a/pkg/cluster/controllers/storage/healthchecker.go
+++ b/pkg/cluster/controllers/storage/healthchecker.go
@@ -43,7 +43,9 @@ func (c *HealthChecker) CheckHealth(ctx context.Context, svc *lcm.ServiceConfig,
 	}
 
 	var res *lcm.CheckResponse
+
 	var err error
+
 	switch checkOpts.StorageDriver {
 	case goharborv1.S3DriverName:
 		res, err = S3StorageHealthCheck(ctx, svc, checkOpts)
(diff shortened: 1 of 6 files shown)


pkg/cluster/controllers/storage/healthchecker.go
```diff
@@ -42,19 +42,29 @@ func (c *HealthChecker) CheckHealth(ctx context.Context, svc *lcm.ServiceConfig,
...
 	}
 
+	var res *lcm.CheckResponse
+
+	var err error
+
 	switch checkOpts.StorageDriver {
 	case goharborv1.S3DriverName:
-		return S3StorageHealthCheck(ctx, svc, checkOpts)
+		res, err = S3StorageHealthCheck(ctx, svc, checkOpts)
```
steven-zou, line 51: @wangcanfengxs 

I suggest wrapping the health check error where the health check func of the specific service health checker is called. Most places are in the `Apply()` method of the service manager. And then you don't need to do changes to the internal logic of the health checker.

e.g: path is `controllers/goharbor/harborcluster/service_mgr.go` 

```go
# line 165
res, err := s.ctrl.HealthChecker().CheckHealth(s.ctx, svcCfg, options...)
		if err != nil {
```

steven-zou (changes requested):
Please consider my suggestion and do changes.

steven-zou:
as we discussed offline, the health check should be refined and then close this PR

PR #429, opened 2021-01-30
add required field in the minio spec for specifying secret

soulseen:
add required field in the minio spec for specifying secret

### DIFF
--- a/pkg/cluster/controllers/storage/minio.go
+++ b/pkg/cluster/controllers/storage/minio.go
@@ -226,6 +226,7 @@ func (m *MinIOController) getMinIOSecretNamespacedName() types.NamespacedName {
 	if secretName == "" {
 		secretName = DefaultPrefix + m.HarborCluster.Name + "-" + DefaultCredsSecret
 	}
+
 	return types.NamespacedName{
 		Namespace: m.HarborCluster.Namespace,
 		Name:      secretName,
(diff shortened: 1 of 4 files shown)


pkg/cluster/controllers/storage/minio.go
```diff
@@ -224,7 +224,7 @@ func (m *MinIOController) getMinIONamespacedName() types.NamespacedName {
 func (m *MinIOController) getMinIOSecretNamespacedName() types.NamespacedName {
 	return types.NamespacedName{
 		Namespace: m.HarborCluster.Namespace,
-		Name:      DefaultPrefix + m.HarborCluster.Name + "-" + DefaultCredsSecret,
+		Name:      m.HarborCluster.Spec.InClusterStorage.MinIOSpec.SecretRef,
```
steven-zou, line 227: @soulseen 

Can we make this field be an optional item? When the user does not provide, we keep the original default behavior that is auto-generating a random secret.

steven-zou (commented):
@soulseen what do u think my advice?

pkg/cluster/controllers/storage/minio.go
soulseen: agree, and updated.

steven-zou (approved):
LGTM

bitsf (approved):

PR #432, opened 2021-02-02
Dev/feat tests

holyhope:


.github/workflows/tests.yml
```diff
@@ -117,7 +117,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: go tests
       run: |
-        make go-test
+        GO_TEST_OPTS='-p 1 -vet=off' make go-test
```
sguyennet, line 120: Does the CI doesn't succeed without this?

holyhope: Checking

sguyennet (approved):

Issue #347, opened 2020-12-17
Replication failed x509 error when redirect is true


chlins:
What can we help you?

![image](https://user-images.githubusercontent.com/31262637/102468726-dfd7e800-408c-11eb-97de-a12bc84b52e5.png)

chlins:
After inject sample-public-certs to  core and jobsevice pod. It woks.

![image](https://user-images.githubusercontent.com/31262637/102973951-a7c32000-4538-11eb-8b0e-ccb664ad0e6d.png)

steven-zou:
- [ ]  Core
- [ ] JS
- [ ] Trivy

Inject ca to the above components

holyhope:
Hello
FYI certmanager project has a csi to inject _properly_ certificates: https://cert-manager.io/v0.16-docs/usage/csi/
WDYT about using it?

bitsf:
@holyhope csi seems apply to handling internal tls creating, while current issue is just need to inject the harbor public cert into some pods' special folder, I'm afraid CSI can't work for it

steven-zou:
And it's also an experimental feature. @holyhope 

But it seems to be useful for distributing Harbor cert to all the nodes where the scheduled pods are running.

PR #443, opened 2021-02-09
feat: inject s3 cert to harbor components(core, jobservice, trivy)

chlins:
Fix: https://github.com/goharbor/harbor-operator/issues/347


heww (approved):

heww:
Should we also need to support `CertificateInjection` in the registry and registryctl components?

steven-zou (approved):
approved

chlins:
> Should we also need to support `CertificateInjection` in the registry and registryctl components?

@heww  Maybe we can add it to other components when needed later.

PR #462, opened 2021-03-02
fix(pg) harbor cluster pg status unknow

Ghostbaby:
harbor cluster pg status unknow


pkg/cluster/controllers/database/readiness.go
```diff
@@ -43,17 +43,28 @@ const (
...
 
-	conn, _, err = p.GetInClusterDatabaseInfo(ctx)
+	conn, client, err = p.GetInClusterDatabaseInfo(ctx)
 	if err != nil {
 		return nil, err
 	}
 
+	defer client.Close(ctx)
+
+	if err := client.Ping(ctx); err != nil {
```
steven-zou, line 60: We should not check the status with ping as the client in the operator process may not connect to psql instance deployed in another namespace.

Why do not use the status of psql CR?

@Ghostbaby

steven-zou (changes requested):
Can we fix it ASAP?

pkg/cluster/controllers/database/readiness.go
Ghostbaby: done

pkg/cluster/controllers/database/readiness.go
```diff
@@ -49,11 +51,25 @@ func (p *PostgreSQLController) Readiness(ctx context.Context) (*lcm.CRStatus, er
 
 	name := p.HarborCluster.Name
 
-	conn, _, err = p.GetInClusterDatabaseInfo(ctx)
+	conn, err = p.GetInClusterDatabaseInfo()
```
steven-zou, line 54: Why do we still need to get conn as we don't check health with that client anymore?

pkg/cluster/controllers/database/readiness.go
```diff
@@ -49,11 +51,25 @@ func (p *PostgreSQLController) Readiness(ctx context.Context) (*lcm.CRStatus, er
...
 	if err != nil {
 		return nil, err
 	}
 
+	var pg api.Postgresql
+	if err := runtime.DefaultUnstructuredConverter.
+		FromUnstructured(p.ActualCR.UnstructuredContent(), &pg); err != nil {
```
steven-zou, line 61: How do you make sure the actual CR is the latest version?

steven-zou (commented):
comments left

pkg/cluster/controllers/database/readiness.go
```diff
@@ -49,11 +50,21 @@ func (p *PostgreSQLController) Readiness(ctx context.Context) (*lcm.CRStatus, er
...
+		return nil, err
+	}
+
+	if pg.Status.PostgresClusterStatus != "Running" {
+		return nil, errors.New("fail to check pg")
```
steven-zou, line 65:
```suggestion
		return nil, errors.New("pg cluster is not ready")
```

steven-zou (commented):
comment left

heww (approved):

steven-zou (approved):
LGTM

PR #498, opened 2021-03-15
feat: watch the crd and owns object after crd ready

heww:
In the first time harbor operator running, the `postgresql` and `redisfailover` CRDs are not created at the harbor operator manager bootstrap stage, harbor cluster controller will be failed when it owns the postgresql and redisfailover. The controller can't get the notification from the status changing of `postgresql` and `redisfailover`.

This PR will watch the CRD and try to own the objects again after a CRD ready.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -1,0 +1,241 @@
+package builder
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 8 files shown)


Ghostbaby (approved):

pkg/builder/builder.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (blder *Builder) WithOptions(options controller.Options) *Builder {
+	blder.blder.WithOptions(options)
+
+	return blder
+}
+
+// TryOwns owns the object when the crdDependency found, otherwise try owns it after the crdDependency found.
+func (blder *Builder) TryOwns(object runtime.Object, crdDependency string, predicates ...predicate.Predicate) *Builder {
```
wangcanfengxs, line 86: Suggest changing the name to TryOwnsCRD, because it only works for CRD

wangcanfengxs (commented):
LGTM

pkg/builder/builder.go
heww: The best name for this method is `TryOwnsWhenCRDReady` but this is too long.

chlins (approved):
LGTM

steven-zou (approved):
LGTM

soulseen (approved):
/lgtm

PR #511, opened 2021-03-22
fix(harborcluster): fix reconcile not work the harborcluster changed.

### LINKED ISSUE #504: switch from harbor core to harbor full stack not success
1. install harbor with manifests/samples/full_stack_fs.yaml
2. update harbor with manifests/samples/full_stack_other_fs.yaml
3. other components like trivy notary chart are not installed, no new deployment created

wangcanfengxs:
use last-applied-hash annotations to check whether harborcr is changed

CLOSE #504


pkg/cluster/controllers/harbor/harbor.go
```diff
@@ -139,6 +138,17 @@ func (harbor *Controller) getHarborCR(harborcluster *v1alpha2.HarborCluster, dep
 	// inject cert to harbor comps
 	injectS3CertToHarborComponents(harborCR)
 
+	harbor.Log.Info("get harbor cr.", "harborcluster", harborcluster, "harborCR", harborCR)
+
+	hash, err := hashstructure.Hash(harborCR, hashstructure.FormatV2, nil)
```
steven-zou, line 143: What about providing an common equality comparing function, like `k8s.HashEqual(o1, o2 runtime.Object) bool`?

steven-zou (commented):
Please see my comments, thanks!

steven-zou:
@wangcanfengxs 

You need to squash your commits in this PR.

steven-zou (approved):
LGTM

steven-zou (changes requested):
@wangcanfengxs

You need to squash your commits in this PR.

pkg/cluster/k8s/annotations.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"strconv"
+
+	"github.com/mitchellh/hashstructure/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// HarborClusterLastAppliedHash contains the last applied hash.
+	HarborClusterLastAppliedHash = "goharbor.io/last-applied-hash"
```
holyhope, line 12: What is the difference with `metadata.generation` and `status.obsrvedGeneration`?
Do we really need the hash of the spec?
Could not easily find official documentation from my phone, but here is a thread to introduce generation and resource version: https://stackoverflow.com/questions/47100389/what-is-the-difference-between-a-resourceversion-and-a-generation

holyhope (commented):
IMO the annotation is unnecessary and the generation data can be used

pkg/cluster/k8s/annotations.go
steven-zou: @wangcanfengxs 

Let's take a look at @holyhope's suggesting solution.

steven-zou: @holyhope 

The implementation shown in this PR is following another discussing solution: https://github.com/kubernetes-sigs/kubebuilder/issues/592#issuecomment-484474923

wangcanfengxs: When I change the annotations or labels of a k8s object, the `generation` will increase.

steven-zou: `Do we really need the hash of the spec?` If not, how can we know the spec is changed？ What we want to compare are the spec of the CR object retrieved from server and the spec regenerated?

@holyhope

steven-zou: @holyhope 

Is this assumption correct? `generation` will be changed if the CR is updated into etcd and `observedGeneration` will only be changed if the CR status is updated by reconciling controller?

some code form the harbor controller:
```go
if !found || generation != observedGeneration {
	return c.preUpdateConditions(ctx, data)
}
```
cc @wangcanfengxs

steven-zou: >generation integer | A sequence number representing a specific generation of the desired state. Populated by the system. Read-only.
-- | --
observedGeneration integer | The generation observed by the deployment controller.
-- | --

steven-zou: a good doc: https://alenkacz.medium.com/kubernetes-operator-best-practices-implementing-observedgeneration-250728868792

steven-zou: More investigation results: so far, the service operators which `harborcluster` is relying on (psql,redis and minio) do not support `observedGeneration` pattern. For those services, maybe we still need such a hash comparison mechanism. The HarborCluster may involve the `observedGeneration` pattern.

steven-zou: k8s API convention: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

holyhope: > When I change the annotations or labels of a k8s object, the `generation` will increase.

This is wrong. The generation will increase only when `spec` field changes.

An issue with computing the hash from local spec is that it will not take care of any mutatingwebhook.
So the hash stored in annotation may not match the spec of the resource.

holyhope: By using the observedGeneration it is possible to use the [StatusClient](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client?utm_source=gopls#StatusClient) and not the full client ensuring the client can not update the generation.

steven-zou: > > When I change the annotations or labels of a k8s object, the `generation` will increase.
> 
> This is wrong. The generation will increase only when `spec` field changes.
> 
> An issue with computing the hash from local spec is that it will not take care of any mutatingwebhook.
> So the hash stored in annotation may not match the spec of the resource.

Is that the case? I think the hash is generated in the same way based on the providing spec.  The comparison does not happen between generated spec and spec retrieved from the server, it happens between the last saved hash and the new hash.

holyhope: As discussed on Thursday, I know understand the need. I approve your great idea using hash of a structure to annotate the dependent resource (hash of the owner sub-field stored as annotation of the owned resource). :-)

This kind of mechanism already exist in [Dependencies.ComputeChecksum()](https://github.com/goharbor/harbor-operator/blob/master/pkg/resources/checksum/checksum.go#L62-L68) used by [Controller.Changed](https://github.com/goharbor/harbor-operator/blob/master/pkg/controller/resource.go#L71).

I suggest to integrate the fix in that package by replacing the `ComputeChecksum()` by a new one computing checksum thankts to the hash of the right field.

WDYT?

steven-zou: Yup, @wangcanfengxs is working on it. And once it's ready, you can review the PR to make sure it's following the way we discussed before.

PR #514, opened 2021-03-22
feat(config) Add dedicated config files for controllers

holyhope:


holyhope:
Goal of this PR is to be able to configure each controller and have a well documented helm values.

pkg/controller/template/template.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, errors.Wrap(err, "path")
+	}
+
+	fileStat, err := os.Stat(configTemplatePath)
+	if err != nil {
+		return nil, errors.Wrap(err, "stat")
+	}
+
+	if fileStat.IsDir() {
+		return nil, errors.Wrap(err, "not a file")
```
XavierDuthil, line 57: `err` is nil
```suggestion
		return nil, fmt.Errorf("%s is not a file", configTemplatePath)
```

pkg/controller/template/template.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (t *Template) RegisterProvider(store *configstore.Store, name string) {
+	store.FileCustomRefresh(t.Path, func(data []byte) ([]configstore.Item, error) {
+		t.refreshed = true
+
+		return []configstore.Item{configstore.NewItem(t.Key, string(data), t.Priority)}, nil
+	})
+}
+
+func (t *Template) ReadyzCheck(req *http.Request) error {
```
XavierDuthil, line 71: Could be renamed to `Ready` and simply return a boolean

pkg/controller/template/template.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if fileStat.IsDir() {
+		return nil, errors.Wrap(err, "not a file")
+	}
+
+	return New(configTemplatePath), nil
+}
+
+func (t *Template) RegisterProvider(store *configstore.Store, name string) {
```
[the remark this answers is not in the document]
XavierDuthil: Why separate this method from `New`, is there a reason to have a template without running this method?

XavierDuthil (commented):
Review - first part

pkg/controller/template/template.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (t *Template) ReadyzCheck(req *http.Request) error {
+	if t.refreshed {
+		return nil
+	}
+
+	return NotYetRefreshed{}
+}
+
+func (t *Template) HealthzCheck(req *http.Request) error {
```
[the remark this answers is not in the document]
holyhope: Thoses methods are used in controllers files:

```golang
	if err := mgr.AddReadyzCheck(r.NormalizeName(ctx, "template"), templateConfig.ReadyzCheck); err != nil {
		return errors.Wrap(err, "cannot add template ready check")
	}

	if err := mgr.AddHealthzCheck(r.NormalizeName(ctx, "template"), templateConfig.HealthzCheck); err != nil {
		return errors.Wrap(err, "cannot add template health check")
	}
```

pkg/controller/template/template.go
holyhope: This should match the [`healthz.Checker` signature](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.6.2/pkg/healthz?utm_source=gopls#Checker)

XavierDuthil: Okay we should probably specify it in a comment then, as there is no other occurrence in the project it's not trivial to figure out.
```suggestion
// Should respect the healthz.Checker signature (from sigs.k8s.io/controller-runtime/pkg/healthz)
func (t *Template) ReadyzCheck(req *http.Request) error {
```

pkg/controller/template/template.go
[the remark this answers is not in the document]
holyhope: `Register()` function take a configstore to update, whereas the `FromConfigStore()` function use configstore to fetch file path.
Those may be 2 different stores, and in order to not get confused with arguments and to not make mandatory having both configstore on init, I decided to split it in 2 methods.

pkg/controller/template/template.go
holyhope: I added the check at compilation using:

```golang
var _ healthz.Checker = (&Template{}).HealthzCheck
```

XavierDuthil:
There is a merge commit in the Pull request, they should be squashed

XavierDuthil (changes requested):

holyhope:
Thank you @XavierDuthil for your review :-)

heww (approved):

bitsf (approved):

PR #526, opened 2021-04-01
Feature/fix checksum

wangcanfengxs:


controllers/goharbor/harborcluster/ctrl_setup.go
```diff
@@ -156,7 +155,7 @@ var harborClusterPredicateFuncs = predicate.Funcs{
 			return true
 		}
 		// when status was not changed and spec was not changed, not need reconcile
-		if equality.Semantic.DeepDerivative(oldObj.Spec, newObj.Spec) && oldObj.Status.Status == newObj.Status.Status {
+		if newObj.Status.ObservedGeneration == newObj.Generation {
```
holyhope, line 158: IMO the comment does not match the condition. The condition means _"Is the new object being reconciled?"_

In order to match the comment, I suggest:
```
If newObj.Generation == oldObj.Generation && equality.Semantic.DeepDerivative(oldObj.Statud, newObj.Status) {
```

WDYT?

pkg/cluster/controllers/harbor/harbor.go
```diff
@@ -57,9 +57,11 @@ func (harbor *Controller) Apply(ctx context.Context, harborcluster *goharborv1.H
 		return harborUnknownStatus(GetHarborCRError, err.Error()), err
 	}
 
+	dep := checksum.New(harbor.Scheme)
+	dep.Add(ctx, harborcluster, true)
+
 	// Found the existing one and check whether it needs to be updated
-	same := equality.Semantic.DeepDerivative(desiredCR.Spec, harborCR.Spec)
-	if !same {
+	if !dep.ChangedFor(ctx, harborCR) {
```
holyhope, line 64: I suggest to write an helper function to make the 3 lines. It seems to be used many times.

holyhope (commented):
Great work!
Thank you a lot!

PR #535, opened 2021-04-06
fix(database) the core and notary pods cannot start up

### LINKED ISSUE #528: The core and notary pods cannot start up when reinstall the Harbor instance
It seems that the secret which contains the database password isn't deleted when removing the Harbor instance, this causes the core and notary pods reporting invalid database password when starting up

Ghostbaby:
The core and notary pods cannot start up when reinstall the Harbor
instance

fix #528

### DIFF
--- a/pkg/cluster/controllers/database/readiness.go
+++ b/pkg/cluster/controllers/database/readiness.go
@@ -59,12 +59,7 @@ func (p *PostgreSQLController) Readiness(ctx context.Context, harborcluster *goh
 		return nil, err
 	}
 
-	secret, err := p.DeployComponentSecret(ctx, conn, harborcluster.Namespace, getDatabasePasswordRefName(name))
-	if err != nil {
-		return nil, err
-	}
-
-	if err := controllerutil.SetControllerReference(harborcluster, secret, p.Scheme); err != nil {
+	if err := p.DeployComponentSecret(ctx, conn, getDatabasePasswordRefName(name), harborcluster); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cluster/controllers/database/readiness.go
```diff
@@ -106,22 +102,27 @@ func getDatabasePasswordRefName(name string) string {
 }
 
 // DeployComponentSecret deploy harbor component database secret.
-func (p *PostgreSQLController) DeployComponentSecret(ctx context.Context, conn *Connect, ns string, secretName string) (*corev1.Secret, error) {
+func (p *PostgreSQLController) DeployComponentSecret(ctx context.Context, conn *Connect, ns string, secretName string,
```
steven-zou, line 105: If the whole `*goharborv1.HarborCluster` has been passed into the func arguments, the namespace arg can be removed, right?  @Ghostbaby

pkg/cluster/controllers/database/readiness.go
```diff
@@ -106,22 +102,27 @@ func getDatabasePasswordRefName(name string) string {
...
+	harborcluster *goharborv1.HarborCluster) error {
+
 	secret := &corev1.Secret{}
 	sc := p.GetDatabaseSecret(conn, ns, secretName)
+	if err := controllerutil.SetControllerReference(harborcluster, sc, p.Scheme); err != nil {
```
steven-zou, line 110: Why let the cluster own the secret? The cluster has owned the psql CR, so is it possible to let the psql CR own the secret?

steven-zou (changes requested):
Please check comments

pkg/cluster/controllers/database/readiness.go
Ghostbaby: yes.

pkg/cluster/controllers/database/readiness.go
Ghostbaby: This secret is store pg conn info for notary and core component, so harborcluster should own this secret.

steven-zou: 👌

steven-zou (approved):
LGTM

heww (approved):

PR #544, opened 2021-04-09
update(storage): update default minio resource limits and requests

soulseen:

### DIFF
--- a/pkg/cluster/controllers/storage/tenant.go
+++ b/pkg/cluster/controllers/storage/tenant.go
@@ -247,19 +247,7 @@ func (m *MinIOController) getResourceRequirements(harborcluster *goharborv1.Harb
 		return &harborcluster.Spec.InClusterStorage.MinIOSpec.Resources
 	}
 
-	limits := map[corev1.ResourceName]resource.Quantity{
-		corev1.ResourceCPU:    resource.MustParse("250m"),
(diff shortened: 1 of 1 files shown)


pkg/cluster/controllers/storage/tenant.go
```diff
@@ -247,19 +247,7 @@ func (m *MinIOController) getResourceRequirements(harborcluster *goharborv1.Harb
...
-	requests := map[corev1.ResourceName]resource.Quantity{
-		corev1.ResourceCPU:    resource.MustParse("250m"),
-		corev1.ResourceMemory: resource.MustParse("512Mi"),
-	}
-
-	return &corev1.ResourceRequirements{
-		Limits:   limits,
-		Requests: requests,
-	}
+	return &corev1.ResourceRequirements{}
```
heww, line 250: Can we just return nil here?

soulseen: I don't sure this is right, but return `&corev1.ResourceRequirements{}` must be correct。

soulseen: looks like return `nil` works well,  and maybe it is a better choice, and to avoid memory escapes.

chlins (approved):
LGTM

bitsf (approved):

PR #547, opened 2021-04-12
fix(harborcluster): improve the get -o printing columns

wangcanfengxs:

### DIFF
--- a/controllers/goharbor/harborcluster/status.go
+++ b/controllers/goharbor/harborcluster/status.go
@@ -50,7 +50,7 @@ func newStatus(source *goharborv1.HarborCluster) *status {
 		cr:     source,
 		locker: &sync.Mutex{},
 		data: &goharborv1.HarborClusterStatus{
-			Status:     goharborv1.StatusUnknown,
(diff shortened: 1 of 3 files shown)


controllers/goharbor/harborcluster/status.go
```diff
@@ -50,7 +50,7 @@ func newStatus(source *goharborv1.HarborCluster) *status {
 		cr:     source,
 		locker: &sync.Mutex{},
 		data: &goharborv1.HarborClusterStatus{
-			Status:     goharborv1.StatusUnknown,
+			Status:     goharborv1.StatusProvisioning,
```
steven-zou, line 53: It seems you did not touch any overall status calculation logic in this PR, is that correct?

steven-zou (commented):
see my comments

controllers/goharbor/harborcluster/status.go
steven-zou: Offline discussed, it's clear now.

steven-zou (approved):
LGTM

heww (approved):

PR #555, opened 2021-04-13
fix: set resources of minio from spec of harbor cluster

heww:

### DIFF
--- a/manifests/samples/harbor_full_minio.yaml
+++ b/manifests/samples/harbor_full_minio.yaml
@@ -130,19 +130,11 @@ spec:
       volumesPerServer: 2
       volumeClaimTemplate:
         spec:
-          storageClassName: standard
           accessModes:
             - ReadWriteOnce
           resources:
             requests:
               storage: 10Gi
-      resources:
-        limits:
-          cpu: 500m
-          memory: 700Mi
-        requests:
-          cpu: 200m
-          memory: 500Mi
   inClusterCache:
     kind: Redis
     redisSpec:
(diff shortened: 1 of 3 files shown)


pkg/cluster/controllers/storage/tenant.go
```diff
@@ -215,7 +215,7 @@ func (m *MinIOController) generateMinIOCR(ctx context.Context, harborcluster *go
 					Servers:             harborcluster.Spec.InClusterStorage.MinIOSpec.Replicas,
 					VolumesPerServer:    harborcluster.Spec.InClusterStorage.MinIOSpec.VolumesPerServer,
 					VolumeClaimTemplate: m.getVolumeClaimTemplate(harborcluster),
-					Resources:           *m.getResourceRequirements(harborcluster),
+					Resources:           harborcluster.Spec.InClusterStorage.MinIOSpec.Resources,
```
Ghostbaby, line 218: we will not define default resource?

soulseen (approved):
/lgtm

manifests/samples/harbor_full_minio.yaml
```diff
@@ -130,19 +130,11 @@ spec:
       volumesPerServer: 2
       volumeClaimTemplate:
         spec:
-          storageClassName: standard
```
bitsf, line 133: should also change in harbor_core_minio.yaml

pkg/cluster/controllers/storage/tenant.go
heww: Why do we need to define the default resource for the minio pod when the customers don't want the limited resources used by it?

steven-zou (approved):
LGTM

holyhope (approved):

PR #557, opened 2021-04-15
clean up some code

bitsf:
clean up some code


.goreleaser.yml
```diff
@@ -16,10 +16,6 @@ builds:
   # Get path to OperatorVersion using: $(go list -m)
   - -X github.com/goharbor/harbor-operator.OperatorVersion={{ if .Tag }}{{ .Tag }}{{ else }}dev{{ end }}
 
-dockers:
-- image_templates:
-  - holyhope/harbor-operator:{{ .Tag }}
```
holyhope, line 21: I suggest that we keep using `dockers`, but to build the `goharbor/harbor-operator` image.
<https://goreleaser.com/customization/docker/>
My goal is to have all release configuration in a single file.
WDYT?

apis/meta/v1alpha1/component.go
```diff
@@ -31,24 +30,6 @@ const (
...
-
-	for i := 0; i < int(componentCount); i++ {
-		size := len(Component(i).String())
-		if size > max {
-			max = size
-		}
-	}
-
-	return max
-}
```
holyhope, line 50: FYI This was used by webhooks to validate `.metadata.name` of custom resources.
Since operator adds suffixes to the name of owned resources, it can reach the native Kubernetes limits: <https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names>.
This validation methods check that child resources will be valid.

pkg/version/version.go
```diff
@@ -9,7 +9,7 @@ const (
 )
 
 var (
-	knowVersions       = []string{}
+	knowVersions       []string
 	knowVersionIndexes = map[string]int{}
```
holyhope, line 13: This make `knownVersions` be `nil`, is that ok?
Can be `knownVersionIndexes` `nil` to?
```suggestion
	knowVersions       []string
	knowVersionIndexes map[string]int
```

holyhope (approved):

apis/meta/v1alpha1/component.go
bitsf: got it

.goreleaser.yml
bitsf: ok

pkg/version/version.go
bitsf: this change is just by my IDE suggestion
```
Empty slice declaration via literal 
 Inspection info: Reports slice declaration with empty literal initializer instead of nil.
An empty slice can be represented by nil or an empty slice literal. The first approach is preferred as it does not lead to memory allocation.
See Code review comments: declaring empty slices.
```

`knowVersionIndexes  map[string]int` is not work

bitsf: this change is just by my IDE suggestion
```
Empty slice declaration via literal 
 Inspection info: Reports slice declaration with empty literal initializer instead of nil.
An empty slice can be represented by nil or an empty slice literal. The first approach is preferred as it does not lead to memory allocation.
See Code review comments: declaring empty slices.
```

`knowVersionIndexes  map[string]int` is not work

steven-zou (approved):
LGTM

PR #558, opened 2021-04-15
fix golanglint goerr113

### LINKED ISSUE #465: fix golint goerr113
What can we help you?

bitsf:
fix #465

### DIFF
--- a/apis/goharbor.io/v1alpha3/harborcluster_webhook.go
+++ b/apis/goharbor.io/v1alpha3/harborcluster_webhook.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	"github.com/goharbor/harbor-operator/pkg/version"
+	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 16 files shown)


apis/goharbor.io/v1alpha3/harborcluster_webhook.go
```diff
@@ -66,7 +67,7 @@ func (hc *HarborCluster) ValidateUpdate(old runtime.Object) error {
 
 	obj, ok := old.(*HarborCluster)
 	if !ok {
-		return fmt.Errorf("failed type assertion on kind: %s", old.GetObjectKind().GroupVersionKind().String())
+		return errors.Errorf("failed type assertion on kind: %s", old.GetObjectKind().GroupVersionKind().String())
```
holyhope, line 70: That way to handle the `goerr113` linter looks like a cheat.
It seems that this linter detect that `fmt.Errorf()` is untyped error that is not a *normal* case.
And it does not detect the `errors.Errorf()` which does the same thing.
Am I wrong? WDYT?

holyhope (approved):
I like that you want to activate that linter.
It is a good first step, but IMO we should no more use the untyped errors factory (no more `fmt.Errorf`but later we should also drop the `errors.Errorf`).

apis/goharbor.io/v1alpha3/harborcluster_webhook.go
bitsf: yes, I just reference the note in https://github.com/Djarvur/go-err113#reports, it tells this package would be ignored.
anyway in future it can be refactor again to define some package level errors and use `%w` in Errorf()

heww (approved):

Issue #568, opened 2021-04-19
Failed to configure the harbor instance using the configmap


ywk253100:
```
2021-04-19T13:06:38.531Z INFO harbor-operator.configuration-controller Reconcile end {"controller": "HarborConfiguration", "resource": "yinw/yinw-http-auth-config-ppwtok", "result": {"Requeue":false,"RequeueAfter":0}, "error": "apply harbor configuration error: response status code is 422, body is {\"errors\":[{\"code\":\"UNPROCESSABLE_ENTITY\",\"message\":\"validation failure list:\\nunsupported media type \\\"application/octet-stream\\\", only [application/json] are allowed\"}]}\n", "updateStatusErr": null}
2021-04-19T13:06:38.532Z ERROR controller Reconciler error {"reconcilerGroup": "", "reconcilerKind": "ConfigMap", "controller": "configmap", "name": "yinw-http-auth-config-ppwtok", "namespace": "yinw", "error": "apply harbor configuration error: response status code is 422, body is {\"errors\":[{\"code\":\"UNPROCESSABLE_ENTITY\",\"message\":\"validation failure list:\\nunsupported media type \\\"application/octet-stream\\\", only [application/json] are allowed\"}]}\n"}
```

PR #578, opened 2021-04-20
fix: set http header when put harbor configuration

chlins:
Fix：https://github.com/goharbor/harbor-operator/issues/568


heww (approved):

steven-zou (approved):
LGTM

PR #587, opened 2021-04-21
feat(ci) fix sample

### LINKED ISSUE #586: make helm-install still require github token
`make helm-install` still require github token set in environment first, which should be optional

bitsf:
close #586
1. make github token optional when make helm-install
2. add default CA to generate certs


steven-zou (approved):
LGTM

heww (approved):

PR #619, opened 2021-04-28
update default tag version to v1.0.0

### LINKED ISSUE #617: Switch the operator image tag in the release-1.0.0 branch to `v1.0.0`
Switch the operator image tag in the release-1.0.0 branch to `v1.0.0`

bitsf:
close #617

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 # Image URL to use all building/pushing image targets
-IMG ?= goharbor/harbor-operator:dev
-RELEASE_VERSION ?= 0.0.0-dev
+IMG ?= goharbor/harbor-operator:v1.0.0
+RELEASE_VERSION ?= v1.0.0
 GIT_COMMIT ?= none
 
 CONFIGURATION_FROM ?= env,file:$(CURDIR)/config-dev.yml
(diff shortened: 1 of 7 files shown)


heww (approved):

Makefile
```diff
@@ -1,7 +1,7 @@
 
 # Image URL to use all building/pushing image targets
-IMG ?= goharbor/harbor-operator:dev
-RELEASE_VERSION ?= 0.0.0-dev
+IMG ?= goharbor/harbor-operator:v1.0.0
```
steven-zou, line 3: Do we have an image with the tag `v1.0.0` now?

bitsf: no, there's no image with tag v1.0.0 now, only v1.0.0-rc1
but I think anyway it would be better than tag dev

Makefile
```diff
@@ -1,7 +1,7 @@
...
+RELEASE_VERSION ?= v1.0.0
```
heww, line 4: `RELEASE_VERSION ?= v1.0.0` => `RELEASE_VERSION ?= 1.0.0` , `v` prefix not needed.

PR #620, opened 2021-04-28
refactor: create the minio bucket by the job

heww:
Create the minio bucket by the job in the target namespace of the harbor
to avoid problem when network is not accessable between operator and
harbor namespaces.


steven-zou (approved):
LGTM

bitsf (approved):

PR #691, opened 2021-06-30
Add s3 multipart copy

### LINKED ISSUE #676: Expose additional configuration options for S3 storage
**Is your feature request related to a problem? Please describe.**
The used implementation for S3 has problems with some S3 backends, like this: https://github.com/goharbor/harbor/issues/12317

The proposed mitigation from that ticket does not work with the harbor operator, because the CRD does not allow setting multipartcopythresholdsize. Patching the ConfigMap for the registry pod does resolve the issue, but that's no sustainable solution.

**Describe the solution you'd like**
I would like to set multipartcopythresholdsize and other S3 settings via the CR.

bitsf:
close #676


cndoit18 (approved):
LGTM

steven-zou (approved):
LGTM

PR #697, opened 2021-07-05
fix(ci) support certmanager 1.4

### LINKED ISSUE #693: cert-manager 1.4.0 not support
can't install harbor successfully when using cert-manager 1.4.0
harborcluster-sample-harbor-harbor-notarysigner-authentication: not ready
image
image

bitsf:
close #693

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -144,7 +144,7 @@ jobs:
         certManager:
         - "1.1.1"
 #        - "1.2.0"
-        - "1.3.1"
+        - "1.4.0"
 
         # https://snapcraft.io/microk8s
         k8sVersion:
@@ -220,6 +220,8 @@ jobs:
           kubectl get all -n ${operatorNamespace}
           exit 1
         fi
+
+        sleep 10
 
     - name: install harbor
       run: |
--- a/controllers/goharbor/harbor/notarysigner.go
+++ b/controllers/goharbor/harbor/notarysigner.go
@@ -82,7 +82,7 @@ func (r *Reconciler) GetNotarySignerCertificateAuthority(ctx context.Context, ha
 			Duration: &metav1.Duration{
 				Duration: duration,
 			},
-			CommonName: r.NormalizeName(ctx, harbor.GetName(), controllers.NotarySigner.String()),
+			CommonName: r.NormalizeName(ctx, harbor.GetName(), controllers.NotarySigner.String(), "CA"),
 			IsCA:       true,
 			Usages: []certv1.KeyUsage{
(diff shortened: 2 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -221,6 +221,8 @@ jobs:
           exit 1
         fi
 
+        sleep 10
```
cndoit18, line 224:
```suggestion
        time kubectl wait -n ${operatorNamespace} pod --selector=control-plane=harbor-operator --timeout=300s
```
It seems possible to use this

bitsf: 前面已经 wait deployment 了，此处怀疑是 harbor-operator 即便都已经部署成功了，webhook 也需要点时间才能生效，所以简单的 sleep

holyhope: Not that easy to understand the response, but I agree with @cndoit18 's suggestion

controllers/goharbor/harbor/notarysigner.go
```diff
@@ -82,7 +82,7 @@ func (r *Reconciler) GetNotarySignerCertificateAuthority(ctx context.Context, ha
 			Duration: &metav1.Duration{
 				Duration: duration,
 			},
-			CommonName: r.NormalizeName(ctx, harbor.GetName(), controllers.NotarySigner.String()),
+			CommonName: r.NormalizeName(ctx, harbor.GetName(), controllers.NotarySigner.String(), "CA"),
```
holyhope, line 85: How does this fix cert-manager v1.4 compatibility?

holyhope (approved):

.github/workflows/tests.yml
steven-zou: > 前面已经 wait deployment 了，此处怀疑是 harbor-operator 即便都已经部署成功了，webhook 也需要点时间才能生效，所以简单的 sleep

@bitsf next time, you should leave an English message.

And will you do change for that part?

controllers/goharbor/harbor/notarysigner.go
bitsf: cert-manger 1.4 doesn't allow using same CN between Cert and CA, so I added a suffix to distinguish them

.github/workflows/tests.yml
bitsf: This `sleep 10` is just to workaround CI failed sometimes recently, because webhook seems not started even we already using `kubectl wait deployment`

cndoit18 (approved):
LGTM

steven-zou (approved):
LGTM

PR #711, opened 2021-07-22
fix: add missing component spec copy for exporter

### LINKED ISSUE #708: Changing number of replicas for exporter does not get honored by the operator
We have deployed harbor with the operator version 1.0.1 and with exporter: {}

Then we changed this to
  exporter:
    replicas: 3

but the number of exporters does not get changed and we still have one.

See above

**Versions:**
- harbor operator version: 1.0.1
- harbor version: 2.2.1
- kubernetes version: 1.20.7

-> any idea why we can not increase the number of exporters using the operator?

heww:
Closes #708


cndoit18 (approved):
LGTM

bitsf (approved):

PR #735, opened 2021-08-05
doc(Development): Remove cert-manager deployment

thcdrt:
Hello,

I think we can remove this cert manager command because it causes issues when running the _run_ Makefile target that tries to launch cert-manager too but on a more recent version. I think we should let the _run_ target handle this.

Signed-off-by: Thomas Coudert thomas.coudert@ovhcloud.com

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -30,15 +30,6 @@ make dev-tools
 ### Services
 
 - [Kubernetes cluster](https://kubernetes.io). You can order a Kubernetes cluster on [ovh.com](https://www.ovh.com/fr/public-cloud/kubernetes/). Then configure your environment to have the right [`Kubeconfig`](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/).
-- [CertManager](https://cert-manager.io/docs/installation/kubernetes/#steps) >= 0.11
-
-  ```bash
-  helm install cert-manager \
-    --namespace cert-manager \
-    --version v0.12.0 \
-    jetstack/cert-manager
-  ```
-
 - Jaeger server.
   
   ```bash


docs/development.md
```diff
@@ -30,15 +30,6 @@ make dev-tools
...
 
 - [Kubernetes cluster](https://kubernetes.io). You can order a Kubernetes cluster on [ovh.com](https://www.ovh.com/fr/public-cloud/kubernetes/). Then configure your environment to have the right [`Kubeconfig`](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/).
-- [CertManager](https://cert-manager.io/docs/installation/kubernetes/#steps) >= 0.11
-
-  ```bash
-  helm install cert-manager \
-    --namespace cert-manager \
-    --version v0.12.0 \
-    jetstack/cert-manager
-  ```
```
cndoit18, line 40: Would it be better to replace it with `kubectl apply -f https://github.com/jetstack/cert-manager/releases/latest/download/cert-manager.yaml`

thcdrt: I don't know if it's really relevant to do it here as it's already started in [Makefile](https://github.com/goharbor/harbor-operator/blob/master/Makefile#L450)

thcdrt: What do you think ?

cndoit18: I'm not quite sure if this area is worth modifying

bitsf (approved):
yes, `make install-dependencies` would also install a cert-manager. this step in doc is useless.

docs/development.md
steven-zou: @thcdrt 

If want to let the developer use makefile to prepare the env, you can delete the section. But it seems we should explicitly mention the developer can use makefile to install prerequisite like cert-manager here. WDYT?

thcdrt: Why not, we could redirect to the certmanager Makefile target, but in this case to be consistent we should talk of nginx ingress, postgresql-operator and redis-operator too, don't you think ?

steven-zou:
@thcdrt @bitsf @cndoit18 

So, does that mean we agree to merge this PR?

cndoit18 (approved):

cndoit18:
> @thcdrt @bitsf @cndoit18
> 
> So, does that mean we agree to merge this PR?


I agree with bitsf's comment

PR #739, opened 2021-08-11
feat(sample) use version v1beta1 in samples

bitsf:


steven-zou (approved):
@bitsf 

It seems you directly override the sample manifest with v1beta1 CRD and then we do not need to change the document about samples, right?

PR #744, opened 2021-08-12
fix: let backend port in notary ingress depend on internal tls

### LINKED ISSUE #736: notary ingress configured for port 443 even with internalTLS enabled false
We configured the following:

```
  expose:
    notary:
      ingress:
        host: notary.customer.com
      tls:
        certificateRef: notary-customer-com

internalTLS:
    enabled: false

```
but the ingress created through the operator has 

```
    http:
      paths:
      - backend:
          service:
            name: harbor-cluster-harbor-harbor-notaryserver
            port:
              number: 443

```
Due to that we get an error 503 when trying to access the service.

Adjusting the ingress for notary manually to port 80 works..

See above

**Versions:**
- harbor operator version: v1.0.1
- harbor version: 2.2.1
- kubernetes version: 1.20.7

-> is this a known defect in the operator that could be fixed?

(we assume that the manual workaround with adjusting the ingress manually is overwritten with every future change... )

heww:
Closes #736


cndoit18 (approved):
LGTM

steven-zou (approved):
LGTM

PR #750, opened 2021-08-17
feat: add network ip families support

heww:
Add network IP families support so that we can deploy the harbor cluster in ipv4 only network.


bitsf (approved):

apis/goharbor.io/v1beta1/chartmuseum_types.go
```diff
@@ -61,6 +61,9 @@ type ChartMuseumSpec struct {
 
 	// +kubebuilder:validation:Required
 	Chart ChartMuseumChartSpec `json:"chart"`
+
+	// +kubebuilder:validation:Optional
+	Network *harbormetav1.Network `json:"network,omitempty"`
```
steven-zou, line 66: As all harbor components need that field, why not put it into the `harbormetav1.ComponentSpec `json:"inline"`

steven-zou (commented):
check my comment?

@heww

apis/goharbor.io/v1beta1/chartmuseum_types.go
heww: The Harbor type includes a HarborComponentsSpec. If the network is added to harbormetav1.ComponentSpec, there will be many networks in the Harbor type. It's confusing.

steven-zou (approved):
LGTM

PR #761, opened 2021-08-27
feat(Portal) Make the deployment of Portal optional. Fix #645.

### LINKED ISSUE #645: Make the deployment of Portal optional
Sometime user do not want to deploy Portal.
At the moment there is no way in the Harbor CR to tell the operator not to deploy Portal.

sguyennet:


steven-zou (approved):
LGTM

PR #775, opened 2021-09-21
fix(Makefile): Set default ingress class when installing nginx-ingres…

thcdrt:
Hello,

We need to set a default ingress class, else harbor and notary ingresses are not working.

Signed-off-by: Thomas Coudert thomas.coudert@ovhcloud.com


sguyennet (approved):
LGTM

cndoit18 (approved):
LGTM

PR #806, opened 2021-11-03
fix minio annotation

### LINKED ISSUE #793: Missing minio ingress annotations
My kubernetes cluster has two ingress controllers and I want the harbor operator to use the non-default ingress controller for all components. In my case this means adding the following annotation to the ingress:
`kubernetes.io/ingress.class: nginx`
The CRD definition from https://github.com/goharbor/harbor-operator/blob/v1.1.1/docs/CRD/custom-resource-definition.md#minio allows for annations.

While the top-level expose will create annotations of the ingress for core and notary via

```
  expose:
    core:
      ingress:
        host: <core-host>
        controller: default
        annotations:
          kubernetes.io/ingress.class: nginx
      tls:
        certificateRef: <core-cert-ref>
    notary:
      ingress:
        host: <notary-host>
        controller: default
        annotations:
          kubernetes.io/ingress.class: nginx
      tls:
        certificateRef: <notary-cert-ref>
```

adding ingress annotations for minio via

 ```
 storage:
    kind: MinIO
    spec:
      minIO:
        operatorVersion: "4.0.6"
        replicas: 2
        secretRef: minio-access-secret
        redirect:
          enable: true
          expose:
            ingress:
              host: <minio-host>
              controller: default
              annotations:
                kubernetes.io/ingress.class: nginx
            tls:
              certificateRef: <minio-cert-ref>
```

will not add the annotations to the ingress.

1. Setup a kubernetes cluster with two ingress controllers
2. Use https://github.com/goharbor/harbor-operator/blob/master/manifests/samples/full_stack.yaml and add a ingress annotation for minio as above
3. Notice that the annotation is not added to the ingress

**Versions:**
Please specify the versions of following systems.

- harbor operator version: 1.1.1
- harbor version: 2.3.3
- kubernetes version: 1.21.3

**Additional context:**

- I am not sure but we might need to implement https://github.com/goharbor/harbor-operator/blob/v1.1.1/pkg/cluster/controllers/storage/tenant.go#L276

### PR DESCRIPTION
**Pull request by bitsf**
close #793
allow add customize annotation for minio ingress

### DIFF
--- a/pkg/cluster/controllers/storage/ingress.go
+++ b/pkg/cluster/controllers/storage/ingress.go
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	goharborv1 "github.com/goharbor/harbor-operator/apis/goharbor.io/v1beta1"
-	"github.com/goharbor/harbor-operator/apis/meta/v1alpha1"
+	harbormetav1 "github.com/goharbor/harbor-operator/apis/meta/v1alpha1"
 	"github.com/goharbor/harbor-operator/pkg/cluster/controllers/common"
 	miniov2 "github.com/goharbor/harbor-operator/pkg/cluster/controllers/storage/minio/apis/minio.min.io/v2"
 	"github.com/goharbor/harbor-operator/pkg/cluster/lcm"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/cluster/controllers/storage/ingress.go [hunk 1]
```diff
@@ -105,31 +105,52 @@ func (m *MinIOController) cleanupIngress(ctx context.Context, harborcluster *goh
 	return minioUnknownStatus(), nil
 }
 
-func (m *MinIOController) generateIngress(ctx context.Context, harborcluster *goharborv1.HarborCluster) *netv1.Ingress { // nolint:funlen
-	var tls []netv1.IngressTLS
-
+func (m *MinIOController) getMinioIngressAnnotations(harborcluster *goharborv1.HarborCluster) map[string]string {
+	istls := false
 	if harborcluster.Spec.Storage.Spec.MinIO.Redirect.Expose != nil &&
```
soulseen, line 110:
```suggestion
	if harborcluster.Spec.Storage.Spec.MinIO.Redirect.Expose != nil
```
```
isEnableExpose := false
if harborcluster.Spec.Storage.Spec.MinIO.Redirect.Expose != nil { sEnableExpose = ture }
```
how about using any variable like 'isEnableExpose' to mark whether `Expose` or not, instead of checking it every time.

soulseen (approved):
/lgtm

steven-zou (approved):
LGTM

heww (approved):

Issue #588, opened 2021-04-21
Change cluster spec to remove optional components but no effects


steven-zou:
The pods of the optional components are still there.

bitsf:
we need to add logic even when some component is not set in CR, to support delete existing sub CR.

PR #808, opened 2021-11-09
feat: add garbagecollector

cndoit18:
fixes: #588

### DIFF
--- a/pkg/controller/common.go
+++ b/pkg/controller/common.go
@@ -19,6 +19,7 @@ import (
 	"github.com/pkg/errors"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -34,6 +35,8 @@ type Controller struct {
 	BaseController controllers.Controller
 	Version        string
 	GitCommit      string
+
+	deletableResources map[schema.GroupVersionKind]struct{}
 
 	ConfigStore *configstore.Store
 	rm          ResourceManager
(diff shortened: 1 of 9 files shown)


pkg/graph/graph.go
```diff
@@ -46,21 +81,78 @@ func (rm *resourceManager) AddResource(ctx context.Context, resource Resource, b
...
+			},
+			GroupVersionKind: gvk,
+		}
+
+		if _, ok := rm.primary[key]; ok {
+			return false, nil
+		}
+
+		rm.primary[key] = res
+	case inlineObject:
```
heww, line 131: This interface only appears here once, we don't need to create an interface type, just use `case interface{GetObject() client.Object}`

cndoit18: thx, updated

pkg/controller/common.go
```diff
@@ -64,6 +70,48 @@ func NewController(ctx context.Context, base controllers.Controller, rm Resource
...
+					},
+				},
+			}
+			if err := c.Client.Create(ctx, sar); err != nil {
+				return err
+			}
+
+			if sar.Status.Allowed {
+				c.deletableGVKs[gv.WithKind(rl.APIResources[i].Kind)] = struct{}{}
+			}
```
bitsf, line 112: how about logging exceptions with `else`

cndoit18: should not be logging exceptions, the effect here is similar to filter

heww (approved):

bitsf (approved):

pkg/graph/graph.go
```diff
@@ -22,6 +22,19 @@ func NewResourceManager() Manager {
 	}
 }
 
+func (rm *resourceManager) GetAllResource(ctx context.Context) []Resource {
```
chlins, line 25: GetAllResources?

pkg/graph/graph.go
```diff
@@ -22,6 +22,19 @@ func NewResourceManager() Manager {
...
+	resources := []Resource{}
+
+	rm.lock.RLock()
+	defer rm.lock.RUnlock()
+
+	for resource := range rm.resources {
+		resources = append(resources, resource)
```
chlins, line 32: why we need to copy them from `rm.resources` to a new slice other than use or return `rm.resources` directly?

chlins (approved):
/lgtm

pkg/graph/graph.go
cndoit18: Because the structure of this field is `map[Resource][]Resource`

pkg/graph/graph.go
cndoit18: Thx, updated

pkg/graph/graph.go
chlins: Thx, got it.

steven-zou (approved):
LGTM

yxxhero (approved):

PR #817, opened 2021-11-17
chore: build controller-gen and stringer in temp dir

heww:
Build the controller-gen and stringer in temp dir to avoid the version of harbor pkg always changing in github actions.


cndoit18 (approved):
LGTM

chlins (approved):
/lgtm

PR #855, opened 2022-03-05
use fixed postgresql/redis image in go tests

bitsf:
CI "Go tests" is failing, that NotaryServer and NotarySigner pods fail to start
This is because notary binary can't connect to lasted postgres version 14.2 with docker image bitnami/postgresql
So use fixed version for postgres and redis


soulseen (approved):
/lgtm

heww (approved):

PR #856, opened 2022-03-05
upgrade default postgresql image to spilo-13

bitsf:


bitsf:
delay to release 1.3, because it is hard to in place upgrade from spilo 1.2 to spilo 1.3 now

Issue #825, opened 2021-12-10
Trivy scanner not removed after remove trivy component.


chlins:
What can we help you?

Deploy `full_stack` harborcluster and the trivy scanner was not removed after removing trivy component, does this behavior by design? 

![image](https://user-images.githubusercontent.com/31262637/145531625-79e7b14c-db56-4271-9e45-23e0dd4f8eb1.png)

bitsf:
there's an env of core pod should also be removed

Issue #824, opened 2021-12-10
Harbor not work after remove chartmuseum component. 


chlins:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
A clear and concise description of what you expected to happen and what's the actual behavior. If applicable, add screenshots to help explain your problem.

Deploy harbor cluster by `full_stack`, and when it all work,  update the cr of harborcluster to remove the chartmuseum component.  Although the pod of Harbor chartmuseum has been removed, but user can not enter to project view in UI and API got 500 error.

**Steps to reproduce the problem:**
Please provide the steps to reproduce this problem.

1. deploy `full_stack`
2. update harbor cluster(remove chartmuseum)
3. check harbor project in ui and core log

**Versions:**
Please specify the versions of following systems.

- harbor operator version: [x.y.z]
- harbor version: [x.y.z]
- kubernetes version: [x.y.z]
- Any additional relevant versions such as CertManager

**Additional context:**

ui screenshot

![image](https://user-images.githubusercontent.com/31262637/145529821-5c78fcad-af58-4e5d-80b0-3d461b96326c.png)

core.log
```
 [ERROR] [/lib/http/error.go:54]: {"errors":[{"code":"UNKNOWN","message":"get chart count of project 1 failed: get content failed: send request GET /library/index.yaml failed: Get \"https://harborcluster-sample-harbor-harbor-chartmuseum/library/index.yaml\": dial tcp: lookup harborcluster-sample-harbor-harbor-chartmuseum on 10.96.0.10:53: server misbehaving"}]}
/harbor/src/server/v2.0/handler/project.go:721, github.com/goharbor/harbor/src/server/v2.0/handler.(*projectAPI).populateProperties
/harbor/src/server/v2.0/handler/project.go:660, github.com/goharbor/harbor/src/server/v2.0/handler.(*projectAPI).getProject
/harbor/src/server/v2.0/handler/project.go:328, github.com/goharbor/harbor/src/server/v2.0/handler.(*projectAPI).GetProject
/harbor/src/server/v2.0/restapi/configure_harbor.go:2129, github.com/goharbor/harbor/src/server/v2.0/restapi.HandlerAPI.func92
/harbor/src/server/v2.0/restapi/operations/project/get_project.go:19, github.com/goharbor/harbor/src/server/v2.0/restapi/operations/project.GetProjectHandlerFunc.Handle
[...]

Issue #743, opened 2021-08-12
apply failed after changing internal tls from false to true


heww:
Steps to reproduce the issue.

1. Deploy the `full_stack.yaml` with internalTLS.enabled=false
2. Change the internalTLS.enabled=true 
3. Apply the new full_stack.yaml

Expect Result
The harbor cluster will be healthy after some minutes

Actual Result
The harbor cluster will be unhealthy forever, the following is the error logs in the harbor operator

```
d4b55069", "request": "cluster-sample-ns/harborcluster-sample-harbor-harbor", "resource.apiVersion": "v1", "resource.kind": "Service", "resource.name": "harborcluster-sample-harbor-harbor-notaryserver", "resource.namespace": "cluster-sample-ns", "key": {"namespace": "cluster-sample-ns", "name": "harborcluster-sample-harbor-harbor-notaryserver"}, "kind": "/v1, Kind=Service", "error": "Service \"harborcluster-sample-harbor-harbor-notaryserver\" is invalid: spec.ports[1].name: Duplicate value: \"api\""}
github.com/goharbor/harbor-operator/pkg/controller.(*Controller).applyAndCheck
        /home/ubuntu/Sync/harbor-operator/pkg/controller/common.go:165
github.com/goharbor/harbor-operator/pkg/controller.(*Controller).ProcessFunc.func1
        /home/ubuntu/Sync/harbor-operator/pkg/controller/resource.go:140
github.com/goharbor/harbor-operator/pkg/graph.(*resourceManager).Run.func1
        /home/ubuntu/Sync/harbor-operator/pkg/graph/runner.go:42
golang.org/x/sync/errgroup.(*Group).Go.func1
        /home/ubuntu/Sync/harbor-operator/vendor/golang.org/x/sync/errgroup/errgroup.go:57
[...]

PR #876, opened 2022-04-12
fix(controller): use update instead of patch

cndoit18:
Signed-off-by: cndoit18 <cndoit18@outlook.com>

Fixes: #825
Fixes: #824
Fixes: #743

### DIFF
--- a/pkg/controller/apply.go
+++ b/pkg/controller/apply.go
@@ -12,8 +12,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var force = true
-
 func (c *Controller) Apply(ctx context.Context, res *Resource) error {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "apply")
 	defer span.Finish()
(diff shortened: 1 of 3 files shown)


pkg/controller/apply.go
```diff
@@ -47,8 +45,7 @@ func (c *Controller) Apply(ctx context.Context, res *Resource) error {
 
 	l.Info("apply changing", "key", key, "kind", resource.GetObjectKind().GroupVersionKind())
 
-	if err := c.Client.Patch(ctx, resource, client.Apply, &client.PatchOptions{
-		Force:        &force,
+	if err := c.Client.Update(ctx, resource, &client.UpdateOptions{
```
steven-zou, line 48: So switch back to update and then append a new mutation to fix the problems caused by using update? @cndoit18

cndoit18: > 

right

bitsf:
we need to do more test this first, the fix is a little tricky.

pkg/controller/resource.go
```diff
@@ -137,10 +138,92 @@ func (c *Controller) ProcessFunc(ctx context.Context, resource runtime.Object, d
...
+		})
+
+		return errors.Wrapf(
+			c.applyAndCheck(ctx, r),
+			"apply %s (%s/%s)", gvk, namespace, name,
+		)
+	}
+}
+
+func needsHCNodePort(svc *corev1.Service) bool {
```
yxxhero, line 209: add unittest for this func?

pkg/controller/resource.go
```diff
@@ -137,10 +138,92 @@ func (c *Controller) ProcessFunc(ctx context.Context, resource runtime.Object, d
...
+	}
+
+	if svc.Spec.ExternalTrafficPolicy != corev1.ServiceExternalTrafficPolicyTypeLocal {
+		return false
+	}
+
+	return true
+}
+
+func needsNodePort(svc *corev1.Service) bool {
```
yxxhero, line 221: add unittest for this func?

yxxhero: add comments will better.

yxxhero (approved):

PR #877, opened 2022-04-12
fix: registry deployment debug port should match the spec

### LINKED ISSUE #875: Registry metrics cannot be scraped with default configuration
If you are reporting a problem, please make sure the following information are provided:

Prometheus should be able to scrape the metrics endpoint of the registry when it's enabled, but Prometheus is unable to connect the metrics endpoint.

Create a new HarborCluster:

```yaml
apiVersion: goharbor.io/v1beta1
kind: HarborCluster
metadata:
  name: harbor
spec:
  version: "2.3.2"
  registry:
    metrics:
      enabled: true
```

Try to scrape the metrics endpoint:

```shell
kubectl port-forward svc/harbor-harbor-harbor-registry 8001
```

**Versions:**
Please specify the versions of following systems.

- harbor operator version: 1.1.1
- harbor version: 2.3.2
- kubernetes version: 1.21
- Any additional relevant versions such as CertManager

**Additional context:**

The problem is present in the latest version as well and is caused by a hardcoded port number [here](https://github.com/goharbor/harbor-operator/blob/master/controllers/goharbor/registry/deployments.go#L56).

While the default port number is 8001 everywhere, the podspec is hardcoded to 5001.

As a workaround, you can set the metrics port manually to 5001 everywhere else:

```yaml
apiVersion: goharbor.io/v1beta1
kind: HarborCluster
metadata:
  name: harbor
spec:
  version: "2.3.2"
  registry:
    metrics:
      enabled: true
      port: 5001
```

However, with 1.1.1 version of the operator this wasn't enough. I had to manually edit the registry resource as well, because the operator didn't update it. Then I had to restart the registry deployment because the config change didn't trigger a rollout. This last one should probably handled as a separate issue.

sagikazarmark:
Fixes #875

### DIFF
--- a/controllers/goharbor/registry/deployments.go
+++ b/controllers/goharbor/registry/deployments.go
@@ -52,8 +52,7 @@ var (
 )
 
 const (
-	apiPort     = 5000 // https://github.com/docker/distribution/blob/749f6afb4572201e3c37325d0ffedb6f32be8950/contrib/compose/docker-compose.yml#L15
-	metricsPort = 5001 // https://github.com/docker/distribution/blob/b12bd4004afc203f1cbd2072317c8fda30b89710/cmd/registry/config-dev.yml#L34
+	apiPort = 5000 // https://github.com/docker/distribution/blob/749f6afb4572201e3c37325d0ffedb6f32be8950/contrib/compose/docker-compose.yml#L15
 	// registry controller port.
 	httpsPort = 8443
 	httpPort  = 8080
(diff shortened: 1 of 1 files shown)


chlins:
@sagikazarmark I don't think change the deployment port is the best solution because for registry the deault metrics port is 5001, but how about change service spec to fix this issue?

sagikazarmark:
@chlins unfortunately, changing the service port is not enough: the registry debug port is configured [here](https://github.com/goharbor/harbor-operator/blob/master/config/config/assets/registry-config.yaml.tmpl#L32) by the CR and the CR default value is 8001, not 5001.

So the alternative is actually changing the service port and the default in all CRs. Even then, it's half a solution because it would work with the defaults, but it wouldn't work if someone changed the debug port through the CR.

So the correct solution here is to use the port from the CR.

I'd agree that it's absolutely unimportant what the container port is, so if there wasn't an option for changing the port in the CR that would totally be fine...but there is.

chlins:
@sagikazarmark Thank you, that's true, the operator has exposed the config for the metrics port of registry, so it seems no way to change it only adjust deployment container port.

chlins (approved):
lgtm

yxxhero (approved):

cndoit18 (approved):
LGTM

Issue #941, opened 2022-08-07
v2.8.0 tag does not match checksum


lengrongfu:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
```shell
$ src git:(main) go mod tidy
go: downloading github.com/docker/distribution v2.8.0+incompatible
verifying github.com/docker/distribution@v2.8.0+incompatible: checksum mismatch
        downloaded: h1:u9vuu6qqG7nN9a735Noed0ahoUm30iipVRlhgh72N0M=
        go.sum:     h1:l9EaZDICImO1ngI+uTifW+ZYvvz7fKISBAKpg+MbWbY=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.

```

**Steps to reproduce the problem:**
Please provide the steps to reproduce this problem.

**Versions:**
Please specify the versions of following systems.

- harbor operator version: [main]
- harbor version: [x.y.z]
- kubernetes version: [x.y.z]
- Any additional relevant versions such as CertManager

**Additional context:**

- **Harbor dependent services:**
  - Context info of postgreSQL
  - Context info of Redis
  - Context info of storage
- **Log files:** Collect logs and attach them here if have.
- **Kubernetes:** How Kubernetes access was provided (what cloud provider, service-account configuration, ...).

lengrongfu:
ref: https://github.com/distribution/distribution/issues/3590
ref: https://github.com/goharbor/harbor/issues/17295

PR #942, opened 2022-08-07
fix: distribution v2.8.0 tag does not match checksum

lengrongfu:
Signed-off-by: lengrongfu <1275177125@qq.com>

Comprehensive Summary of your change

Upgrade distribution version from v2.8.0 to v2.8.1, because v2.8.0 tag does not match checksum.

Issue being fixed
issue: https://github.com/goharbor/harbor-operator/issues/941

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,7 +58,7 @@ require (
 	github.com/cenkalti/backoff/v4 v4.1.1 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/docker/distribution v2.8.0+incompatible // indirect
+	github.com/docker/distribution v2.8.1+incompatible // indirect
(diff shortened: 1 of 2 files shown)


yxxhero (approved):

go.mod
```diff
@@ -58,7 +58,7 @@ require (
 	github.com/cenkalti/backoff/v4 v4.1.1 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/docker/distribution v2.8.0+incompatible // indirect
+	github.com/docker/distribution v2.8.1+incompatible // indirect
```
bitsf, line 61: the indirect library should not be edit directly.

bitsf:
this distribution would be upgraded when we support harbor 2.6 later

cndoit18 (dismissed):

heww:
We should never update the indirect library manually.

Issue #955, opened 2022-08-27
It is recommended that the `expost.core.ingress` field be changed to optional


lengrongfu:
What can we help you?

Currently `ExternalURL` and `expost.core.ingress` fields are required, can they be changed to not required?
Currently we have a scenario where we only need to access the `Harbor` backend service, and it is accessed inside the `k8s` cluster, so it can be accessed through the `service ClusterIp type` of `harbor core`.

heww:
`ExternalURL` is a required configuration for the harbor-core because the harbor-core uses this configuration to generate the auth service URL.

ingress can be changed to not required

thcdrt:
Fixed in #975

PR #974, opened 2022-11-04
fix: make expose core ingress optional

ShouEnHsiao:
This is to make expose core ingress optional to accommodate the requirement mentioned in  https://github.com/goharbor/harbor-operator/issues/955

Signed-off-by: Shou-En Hsiao <shouenhsiao@google.com>


ShouEnHsiao:
Still debugging the chart test failure. Will push another commit once resolve the issue.

PR #983, opened 2022-11-21
Add github action to assign maintainers to PRs

MarcelMue:
Signed-off-by: Marcel Mueller <marcel.mueller1@rwth-aachen.de>

### DIFF
--- a/.github/auto-assignees.yaml
+++ b/.github/auto-assignees.yaml
@@ -19,6 +19,7 @@ reviewers:
   - cndoit18
   - lubronzhan
   - yxxhero
+  - wy65701436
 
 options:
   ignore_draft: true
--- a/.github/auto-assignees.yaml
+++ b/.github/auto-assignees.yaml
@@ -1,0 +1,30 @@
+# This assigns a PR to its author
+addAssignees: true
+
+addReviewers: false
+
+reviewers:
+  - holyhope
+  - sguyennet
+  - jMonsinjon
+  - steven-zou
+  - mhurtrel
+  - heww
+  - bitsf
+  - wangcanfengxs
+  - Ghostbaby
+  - soulseen
+  - chlins
+  - XavierDuthil
+  - cndoit18
+  - lubronzhan
+  - yxxhero
+
+options:
+  ignore_draft: true
+  ignored_keywords:
+    - WIP
+    - wip
+    - DO NOT MERGE
+  enable_group_assignment: true
+  number_of_reviewers: 3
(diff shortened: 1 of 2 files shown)


cndoit18 (approved):

OrlinVasilev (approved):
looks great to me :)

.github/auto-assignees.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - yxxhero
+
+options:
+  ignore_draft: true
+  ignored_keywords:
+    - WIP
+    - wip
+    - DO NOT MERGE
+  enable_group_assignment: true
+  number_of_reviewers: 3
```
bitsf, line 30: I think 2 is enough

.github/auto-assignees.yaml
```diff
@@ -0,0 +1,30 @@
+# This assigns a PR to its author
+addAssignees: true
```
bitsf, line 2: this seems not required currently

MarcelMue: Why do you think so? Every example (even the upstream one) explicitly sets this to true:

https://github.com/kentaro-m/auto-assign-action#single-reviewers-list

And matches what `harbor` repo has set:
https://github.com/goharbor/harbor/blob/a826720cc94dd9ac8d3495ba74a913055a2a9268/.github/auto-assignees.yml#L3

.github/auto-assignees.yaml
MarcelMue: Changed.

.github/auto-assignees.yaml
OrlinVasilev: it should be either true or author, we need true so we have array of assignees

bitsf: please add wy65701436 , and assign to him

MarcelMue: Do you mean adding @wy65701436 to the list of assignees?

I took the current list from here: https://github.com/goharbor/harbor-operator/blob/master/MAINTAINERS

OrlinVasilev: @MarcelMue correct but Yan Wang @wy65701436 is currently the overall tech lead for Harbor so it's good to have him :)

OrlinVasilev: @wy65701436 can you confirm please

MarcelMue: Done.

OrlinVasilev:
I think that's is done shall we merge ? @goharbor/all-maintainers

Vad1mo (approved):

Issue #638, opened 2021-05-11
Support backend storage driver: Aliyun OSS.


steven-zou:
See https://www.alibabacloud.com/product/oss

PR #991, opened 2022-12-01
support aliyun oss storage

bitsf:
close https://github.com/goharbor/harbor-operator/issues/638

testing reference
```
---
apiVersion: v1
kind: Secret
metadata:
  name: aliyun-access-secret
  namespace: cluster-sample-ns
data:
  secret: <base64 encoded access key>
type: Opaque
---
...
spec:
  storage:
    kind: Oss
    spec:
      redirect:
        enable: true
      oss:
        accessKeyID: <access id>
        accessSecretRef: aliyun-access-secret
        bucketName: harbor-operator
        pathPrefix: test-harbor
        endpoint: oss-us-west-1.aliyuncs.com
        internal: false
        encrypt: true
        chunksize: 10000000
```


controllers/goharbor/chartmuseum/deployments.go
```diff
@@ -124,6 +124,40 @@ func (r *Reconciler) GetDeployment(ctx context.Context, chartMuseum *goharborv1.
...
+		}, corev1.EnvVar{
+			Name:  "STORAGE_ALIBABA_ENDPOINT",
+			Value: chartMuseum.Spec.Chart.Storage.AliOSS.Endpoint,
+		}, corev1.EnvVar{
+			Name:  "ALIBABA_CLOUD_ACCESS_KEY_ID",
+			Value: chartMuseum.Spec.Chart.Storage.AliOSS.AccessKeyID,
+		}, corev1.EnvVar{
+			Name:  "STORAGE_ALIBABA_PREFIX",
+			Value: getChartFolder(chartMuseum.Spec.Chart.Storage.AliOSS.PathPrefix),
+		})
```
MarcelMue, line 144: Why is `STORAGE_ALIBABA_SSE` not included?
https://github.com/helm/chartmuseum/blob/main/pkg/config/vars.go#L463

Seems like it would be a nice setting to expose. Or are there reasons why it's missing here?

apis/goharbor.io/v1beta1/harbor_types.go
```diff
@@ -509,6 +509,11 @@ type HarborStorageImageChartStorageSpec struct {
 	// An implementation of the storagedriver.StorageDriver interface which uses Google Cloud for object storage.
 	// See https://docs.docker.com/registry/storage-drivers/gcs/
 	Gcs *HarborStorageImageChartStorageGcsSpec `json:"gcs,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// An implementation of the storagedriver.StorageDriver interface which uses Google Cloud for object storage.
```
thcdrt, line 514:
```suggestion
	// An implementation of the storagedriver.StorageDriver interface which uses Alibaba for object storage.
```

apis/goharbor.io/v1beta1/harbor_types.go
```diff
@@ -509,6 +509,11 @@ type HarborStorageImageChartStorageSpec struct {
...
+	// See https://docs.docker.com/registry/storage-drivers/oss/
+	AliOSS *HarborStorageImageChartStorageAliOSSSpec `json:"aliOss,omitempty"`
```
thcdrt, line 516: Naming question: we generally use the same name as docker distribution to name storage drivers. In this case it would be oss.
Why prefixing it by Ali ?

apis/goharbor.io/v1beta1/registry_types.go
```diff
@@ -617,6 +617,11 @@ type RegistryStorageDriverSpec struct {
 	// An implementation of the storagedriver.StorageDriver interface which uses Google Cloud for object storage.
 	// https://docs.docker.com/registry/storage-drivers/gcs/
 	Gcs *RegistryStorageDriverGcsSpec `json:"gcs,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// An implementation of the storagedriver.StorageDriver interface which uses Google Cloud for object storage.
```
thcdrt, line 622:
```suggestion
	// An implementation of the storagedriver.StorageDriver interface which uses Alibaba for object storage.
```

config/config/assets/registry-config.yaml.tmpl
```diff
@@ -145,6 +145,16 @@ storage:
     realm: {{ .Spec.Storage.Driver.Azure.BaseURL | quote }}
   {{- end }}
 
+  {{- if .Spec.Storage.Driver.AliOSS }}
+  oss:
```
thcdrt, line 149: According to docker distribution documentation, a lot more options are available like chunksize, endpoint, encryption options.

manifests/harbor/deployment.yaml
```diff
@@ -14776,6 +14814,27 @@ spec:
                 type: string
               imageChartStorage:
                 properties:
+                  aliOss:
+                    description: An implementation of the storagedriver.StorageDriver
+                      interface which uses Google Cloud for object storage. See https://docs.docker.com/registry/storage-drivers/oss/
```
thcdrt, line 14819:
```suggestion
                      interface which uses Alibaba Cloud for object storage. See https://docs.docker.com/registry/storage-drivers/oss/
```

manifests/cluster/deployment.yaml
```diff
@@ -43173,6 +43232,27 @@ spec:
                     type: object
                   driver:
                     properties:
+                      aliOSS:
+                        description: An implementation of the storagedriver.StorageDriver
+                          interface which uses Google Cloud for object storage. https://docs.docker.com/registry/storage-drivers/oss/
```
thcdrt, line 43237:
```suggestion
                          interface which uses Alibaba Cloud for object storage. https://docs.docker.com/registry/storage-drivers/oss/
```

manifests/cluster/deployment.yaml
```diff
@@ -14776,6 +14814,27 @@ spec:
                 type: string
               imageChartStorage:
                 properties:
+                  aliOss:
+                    description: An implementation of the storagedriver.StorageDriver
+                      interface which uses Google Cloud for object storage. See https://docs.docker.com/registry/storage-drivers/oss/
```
thcdrt, line 14819:
```suggestion
                      interface which uses Alibaba Cloud for object storage. See https://docs.docker.com/registry/storage-drivers/oss/
```

manifests/harbor/deployment.yaml
```diff
@@ -30773,6 +30832,27 @@ spec:
                     type: object
                   driver:
                     properties:
+                      aliOSS:
+                        description: An implementation of the storagedriver.StorageDriver
+                          interface which uses Google Cloud for object storage. https://docs.docker.com/registry/storage-drivers/oss/
```
thcdrt, line 30837:
```suggestion
                          interface which uses Alibaba Cloud for object storage. https://docs.docker.com/registry/storage-drivers/oss/
```

controllers/goharbor/chartmuseum/deployments.go
bitsf: because harbor-helm doesn't implement this.
https://github.com/goharbor/harbor-helm/blob/4f622fb27b6dde783c81aa99bae3d59cec6b400b/templates/chartmuseum/chartmuseum-cm.yaml#L107

bitsf: and registry storage setting doesn't have this setting too.
https://docs.docker.com/registry/storage-drivers/oss/
So I think this feature can be considered to add in future release if needed.

apis/goharbor.io/v1beta1/harbor_types.go
bitsf: well, `oss` seems too general, so I add Ali.
But if this is traditional, I will fix this.

config/config/assets/registry-config.yaml.tmpl
bitsf: there would be some confuse when using `endpoint` between registry and chartmuseum, the format are different

apis/goharbor.io/v1beta1/harborcluster_types.go
```diff
@@ -237,6 +238,8 @@ type StorageSpec struct {
 	Azure *AzureSpec `json:"azure,omitempty"`
 	// +kubebuilder:validation:Optional
 	Gcs *GcsSpec `json:"gcs,omitempty"`
+	// +kubebuilder:validation:Optional
+	OSS *OssSpec `json:"oss,omitempty"`
```
thcdrt, line 242:
```suggestion
	Oss *OssSpec `json:"oss,omitempty"`
```

charts/harbor-operator/templates/crds.yaml
```diff
@@ -4116,6 +4116,28 @@ spec:
...
+                            type: string
+                          accessSecretRef:
+                            type: string
+                          bucket:
+                            type: string
+                          endpoint:
+                            type: string
+                          pathPrefix:
+                            type: string
+                          serverSideEncryption:
```
thcdrt, line 4131: Why is it defined inside chartmuseum CRD while it seems that chartmuseum doesn't use it

config/config/assets/registry-config.yaml.tmpl
```diff
@@ -145,6 +145,31 @@ storage:
     realm: {{ .Spec.Storage.Driver.Azure.BaseURL | quote }}
   {{- end }}
 
+  {{- if .Spec.Storage.Driver.Oss }}
+  oss:
+    accesskeyid: {{ .Spec.Storage.Driver.Oss.AccessKeyID | quote }}
+    bucket: {{ .Spec.Storage.Driver.Oss.Bucket | quote }}
+    region: {{ .Spec.Storage.Driver.Oss.Region | quote }}
+    {{- if .Spec.Storage.Driver.Oss.Endpoint }}
+    endpoint: {{ .Spec.Storage.Driver.Oss.Bucket }}.{{ .Spec.Storage.Driver.Oss.Endpoint }}
```
thcdrt, line 154: Endpoint needs to contain bucket ?

apis/goharbor.io/v1beta1/harborcluster_types.go
bitsf: why this is desired? I feel this is similar as `Gcs` which is also a abbreviation.

manifests/cluster/deployment.yaml
```diff
@@ -4126,6 +4126,28 @@ spec:
...
+                            type: string
+                          accessSecretRef:
+                            type: string
+                          bucket:
+                            type: string
+                          endpoint:
+                            type: string
+                          pathPrefix:
+                            type: string
+                          serverSideEncryption:
```
thcdrt, line 4141: Why is it defined inside chartmuseum CRD while it seems that chartmuseum doesn't use it ?

manifests/harbor/deployment.yaml
```diff
@@ -4126,6 +4126,28 @@ spec:
...
+                            type: string
+                          accessSecretRef:
+                            type: string
+                          bucket:
+                            type: string
+                          endpoint:
+                            type: string
+                          pathPrefix:
+                            type: string
+                          serverSideEncryption:
```
thcdrt, line 4141: Why is it defined inside chartmuseum CRD while it seems that chartmuseum doesn't use it

config/config/assets/registry-config.yaml.tmpl
bitsf: I still fix this, as it may be useful for user to use internal endpoint

controllers/goharbor/chartmuseum/deployments.go
bitsf: I just add ServerSideEncryption in ChartMuseumChartStorageDriverOssSpec

manifests/cluster/deployment.yaml
bitsf: yes it is not used at all, just refer what ChartMuseumChartStorageDriverAmazonSpec did, maybe for future use.

config/config/assets/registry-config.yaml.tmpl
bitsf: yes, this is a strange behavior defined by aliyun
image

manifests/cluster/deployment.yaml
thcdrt: Sorry, I don't understand, why do we define it if we don't use it ? It's only defined for amazon, not for gcs...

config/config/assets/registry-config.yaml.tmpl
bitsf: also refer https://github.com/goharbor/harbor-helm/blob/185354a01bc00fb9db5943dfd331128fbcd149d8/templates/registry/registry-cm.yaml#L138

thcdrt (approved):
LGTM

heww (approved):

PR #1002, opened 2023-01-02
Update makefile and change postgres password key from 'postgresql-password' to 'postgres-password'

MarcelMue:
Breaking change: Password key in postgres secret has changed.

Signed-off-by: Marcel Mueller <marcel@giantswarm.io>


.github/workflows/tests.yml
```diff
@@ -201,11 +201,13 @@ jobs:
...
     - name: install harbor-operator
       run: |
+        make kustomize
+
         cd manifests/cluster
         kustomize edit add secret github-token --disableNameSuffixHash --from-literal=GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
         kustomize edit add patch --path patch/github-token.yaml
         kustomize edit set image goharbor/harbor-operator=${dockerImage}
-        kustomize build | kubectl create -f -
+        ../../bin/kustomize build --reorder legacy | kubectl create -f -
```
MarcelMue, line 210: Necessary - we previously used `kustomize` shipped with ubuntu `latest` it seems. That caused us to hit deprecations in the current setup and therefore things were applied in the wrong order. Using kustomize from the make file fixed it.

.github/workflows/tests.yml
```diff
@@ -369,7 +371,7 @@ jobs:
 
       - uses: azure/setup-kubectl@v1
         with:
-          version: 'latest'
+          version: 'v1.25.4'
```
MarcelMue, line 374: I mostly did this to have a predictable kubectl version. Not entirely necessary.

Makefile
```diff
@@ -139,7 +139,7 @@ helm-postgres-operator: helm $(CHARTS_DIRECTORY)/postgres-operator/values.yaml
 	$(MAKE) kube-namespace
 	$(HELM) repo add zalando https://opensource.zalando.com/postgres-operator/charts/postgres-operator
 	$(HELM) repo update
-	$(HELM) upgrade --namespace "$(NAMESPACE)" --install postgres-operator zalando/postgres-operator --version 1.6.3 -f $(CHARTS_DIRECTORY)/postgres-operator/values.yaml
+	$(HELM) upgrade --namespace "$(NAMESPACE)" --install postgres-operator zalando/postgres-operator --version 1.7.0 -f $(CHARTS_DIRECTORY)/postgres-operator/values.yaml
```
MarcelMue, line 142: Needed as 1.6.3 does not exist anymore.

Makefile
```diff
@@ -451,16 +451,17 @@ install-dependencies: certmanager postgresql redis ingress
 .PHONY: redis
 redis: helm sample-redis
 	$(HELM) repo add bitnami https://charts.bitnami.com/bitnami
-	$(HELM) upgrade --install harbor-redis bitnami/redis --version 15.7.0 \
+	$(HELM) upgrade --install harbor-redis bitnami/redis --version 16.13.2 \
```
MarcelMue, line 454: Needed as 15.7.0 does not exist anymore.

Makefile
```diff
@@ -451,16 +451,17 @@ install-dependencies: certmanager postgresql redis ingress
...
 		--set-string auth.existingSecret=harbor-redis \
 		--set-string auth.existingSecretPasswordKey=redis-password
 
 .PHONY: postgresql
 postgresql: helm sample-database
 	$(HELM) repo add bitnami https://charts.bitnami.com/bitnami
-	$(HELM) upgrade --install harbor-database bitnami/postgresql --version 10.14.3 \
-		--set-string initdbScriptsConfigMap=harbor-init-db \
-		--set-string existingSecret=harbor-database-password
+	$(HELM) upgrade --install harbor-database bitnami/postgresql --version 11.6.26 \
```
MarcelMue, line 461: Parameters changed in the upstream chart. Old version doesn't exist anymore.

Makefile
```diff
@@ -554,8 +555,8 @@ $(CONTROLLER_GEN):
 
 # find or download markdownlint
 # download markdownlint if necessary
-MARKDOWNLINT_VERSION := 0.32.2
-MARKDOWNLINT := $(BIN)/markdownlint
+MARKDOWNLINT_VERSION := 0.33.0
+MARKDOWNLINT := npx markdownlint
```
MarcelMue, line 559: Needed as `npm bin` does not exist anymore on the latest ubuntu.

apis/meta/v1alpha1/secrets.go
```diff
@@ -45,7 +45,7 @@ const (
 	SecretTypePostgresql corev1.SecretType = "goharbor.io/postgresql"
 
 	// PostgresqlPasswordKey is the password to connect to postgresql.
-	PostgresqlPasswordKey = "postgresql-password" //nolint:gosec
+	PostgresqlPasswordKey = "postgres-password"
```
MarcelMue, line 48: This changed in the upstream `postgresql` bitnami chart. I fear this is a breaking change for upgrades.

config/samples/database/kustomization.yaml
```diff
@@ -14,5 +14,5 @@ configMapGenerator:
 secretGenerator:
 - name: harbor-database-password
   literals:
-  - postgresql-password=the-database-password
+  - postgres-password=the-database-password
```
MarcelMue, line 17: Same as above - changed because of bitname update.

MarcelMue (commented):
Reasoning for most of the changes. This took a long time because each problem was not obvious to find.

.github/workflows/tests.yml
thcdrt: Yes, that could help to avoid issues but we'll need to think to upgrade it when supporting new Kubernetes versions

MarcelMue: I will restructure this in a follow up PR, so we use the same kubectl version everywhere in CI

thcdrt (commented):
LGTM to me, could you squash some commits ?

thcdrt (approved):
LGTM, thank you !

.github/workflows/tests.yml
Vad1mo: Why do we need **--reorder legacy**?

MarcelMue: It's used consistently through the makefile in this repo. The issues I encountered were definitely because of kustomize, so I aligned it with the rest of the repo ordering.

See here for example: https://github.com/goharbor/harbor-operator/blob/9c19b151ff71f66f9079236e4dc4ce3d4a1e02c3/Makefile#L444

I will file a follow up issue to remove it everywhere.

MarcelMue: Issue: https://github.com/goharbor/harbor-operator/issues/1021

Vad1mo (approved):

.github/workflows/tests.yml
MarcelMue: I added a follow up issue here: https://github.com/goharbor/harbor-operator/issues/1022

PR #1013, opened 2023-02-06
fix: postgresql not deleted with harborcluster (1.4.0)

samuelchen:
add ownerReference to expectCR in PostgreSQLController.Update before comparing. Origin expectCR have no OwnerReference .


MarcelMue:
It currently makes no sense to port this specific change to the 1.4 branch. We are unable to replicate the issue and it's not a blocker for 1.4 for us currently.

PR #1014, opened 2023-02-06
fix: postgresql not deleted with harborcluster (master)

samuelchen:
add ownerReference to expectCR in PostgreSQLController.
Update before comparing. Origin expectCR have no OwnerReference .


cndoit18 (approved):
lgtm

PR #1016, opened 2023-02-13
fix: for update strategy add propagation from cluster to components

karaguo:
1. add `UpdateStrategyType` at registry type
2. set the registry `UpdateStrategyType` value from harborcluster/harbor CRs
3. send the value to registry/registryctl deployment


MarcelMue:
Please sign your commit - see here for reference:
https://github.com/goharbor/harbor-operator/pull/1016/checks?check_run_id=11315256541

karaguo:
> Please sign your commit - see here for reference: https://github.com/goharbor/harbor-operator/pull/1016/checks?check_run_id=11315256541

Thanks! The commit is signed and DCO is passed

thcdrt:
Shouldn't we propagate it on others Harbor components too ? (core, chartmuseum...)

LiuShuaiyi:
> Shouldn't we propagate it on others Harbor components too ? (core, chartmuseum...)

If you were asking if users should be able to configure the UpdateStrategy for the deployments of other components? Only if the components will attach PVC. Not sure if the context is clear about this PR, but https://stackoverflow.com/questions/54322349/updating-a-deployment-that-uses-a-readwriteonce-volume-will-fail-on-mount is essentially the problem that we were faced with when we tried to upgrade the registry component.

LiuShuaiyi:
See https://github.com/goharbor/harbor-helm/blob/master/values.yaml#L349

LiuShuaiyi:
Yeah aligning with harbor-helm's implementation sounds good to me

karaguo:
Thanks @LiuShuaiyi @thcdrt! To align with harbor helm, I added the `UpdateStrategyType` to jobservice too. Now it would take effect on deployments of registry and jobservice.

thcdrt (approved):
Thanks, LGTM

MarcelMue (approved):

OrlinVasilev:
Hiiii @karaguo congrats on merging your first PR here :)) Welcome to the club!!!! :)

PR #1037, opened 2023-05-23
Rework auto assignees to have review groups

MarcelMue:


.github/auto-assignees.yaml
```diff
@@ -1,33 +1,34 @@
...
+    - lubronzhan
+    - yxxhero
+    - wy65701436
+
+skipKeywords:
+  - WIP
+  - wip
+  - DO NOT MERGE
+
+numberOfReviewers: 2
```
thcdrt, line 34:
```suggestion
numberOfReviewers: 1
```

?

thcdrt (changes requested):

.github/auto-assignees.yaml
MarcelMue: The 2 is because:
I added two groups of reviewers - one with the two of us and one with all the other maintainers. The 2 here will cause automation to (hopefully) always assign one of us and one of all the other maintainers as reviewers.

thcdrt: Ho ok I thought it was defining the number of minimal approves, mb

PR #1040, opened 2023-05-23
Remove CRDs from helm completely

MarcelMue:
This is basically a sledgehammer fix to the issue. It completely removes the CRDs from helm and hardcodes the names which we use in our tests.

As you can see this works, but is not a sustainable solution. I will create a follow up issue to come up with a long term solution before release 1.4.


.github/workflows/tests.yml
```diff
@@ -628,6 +628,10 @@ jobs:
       - name: install harbor-operator
         run: |
           set -ex
+          make kustomize
+
+          ./bin/kustomize build --reorder legacy config/helm/crds | kubectl create -f -
```
thcdrt, line 633: Shouldn't we add a Makefile target to install crds ?

config/helm/crds/kustomization.yaml
```diff
@@ -1,12 +1,12 @@
-namePrefix: '{{ include "chart.fullname" . }}-'
+namePrefix: 'harbor-operator-'
 
-namespace: '{{ .Release.Namespace }}'
+namespace: 'harbor-operator-ns'
```
thcdrt, line 3: So we can't install CRDs in other namespace than harbor-operator-ns ?

thcdrt (approved):

.github/workflows/tests.yml
MarcelMue: This is how it is done in the other tests currently - I am happy to create an issue to align this and not do this crazy stuff anymore :D

MarcelMue:
I created #1041 as a follow up.

PR #1047, opened 2023-06-03
docs: add yanwang to the maintainers list

chlins:
Add Yan Wang to the maintainers list.

### DIFF
--- a/.github/auto-assignees.yaml
+++ b/.github/auto-assignees.yaml
@@ -9,6 +9,8 @@ reviewGroups:
     - MarcelMue
     - thcdrt
     - nicolaslacroux
+    - wy65701436
+    - chlins
   general:
     - holyhope
     - steven-zou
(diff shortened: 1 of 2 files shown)


thcdrt (approved):

.github/auto-assignees.yaml
```diff
@@ -20,4 +22,4 @@ skipKeywords:
   - wip
   - DO NOT MERGE
 
-numberOfReviewers: 1
+numberOfReviewers: 2
```
MarcelMue, line 25: I don't think this should be changed. `1` means that one reviewer from each group will be chosen (2 groups, so 2 reviewers). Increasing this to 2 will mean 4 reviewers are pinged from each PR.

I think you want to increase the number of approvers needed to merge a PR, that needs to happen through the project settings itself.

MarcelMue (changes requested):

.github/auto-assignees.yaml
chlins: thx, updated.

MarcelMue (approved):

PR #1062, opened 2023-07-25
chore: update the github actions

chlins:
Update the actions rule pattern from master to main.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master, 0.5.0, 0.5.1, 0.5.2, release-* ]
+    branches: [ main, 0.5.0, 0.5.1, 0.5.2, release-* ]
   pull_request:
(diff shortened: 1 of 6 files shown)


MarcelMue (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master, 0.5.0, 0.5.1, 0.5.2, release-* ]
+    branches: [ main, 0.5.0, 0.5.1, 0.5.2, release-* ]
```
thcdrt, line 16: Can't we remove `0.5.0, 0.5.1, 0.5.2`?

thcdrt (approved):

.github/workflows/codeql-analysis.yml
chlins: At present, I think these branch could be removed, but this PR only aims to update the main branch, let's do the cleanup work not limited to this place in other PR.

PR #1067, opened 2023-08-18
Add GenerationChangedPredicate event filter to project controller

ChristianLoewel:
This pull request removes the `LastApplyTime` status field in the HarborProject CRD.

I have noticed that updating the status field asynchronously after returning the Reconcile function seems to trigger an immediate 're-reconciliation'. As `LastApplyTime` gets updated every time, this results in an endless loop instead of reconciling HarborProjects only every 5 minutes as intended.


chlins:
It's the way kubernetes works, the controller will receive the event when a CR changed but maybe we can retain this field as it's helpful when debug issues, how about update the strategy like add the predicate filter or do the time window check(e.g `if now()-lastAppliedTime < 5min`) before reconcile?

ChristianLoewel:
> It's the way kubernetes works, the controller will receive the event when a CR changed but maybe we can retain this field as it's helpful when debug issues, how about update the strategy like add the predicate filter or do the time window check(e.g `if now()-lastAppliedTime < 5min`) before reconcile?

Yeah, I totally agree with that. Will take a look at the options you mentioned.

chlins (approved):
lgtm

MarcelMue (approved):

PR #1088, opened 2024-09-05
bump Go version to 1.22

yxxhero:


Vad1mo:
@yxxhero somehow all tests fail,

yxxhero:
@Vad1mo go generate issue. I'm finding the key.

Vad1mo:
I also think we should only support and test against the latest k8s releases according to https://endoflife.date/kubernetes

